PR #1452, opened 2018-05-10
Correct some typos and omissions in host organism test data

jseager7:
It's only test data, but there was a typo on 'tobacco', and GitHub pointed out that _Rattus rattus_ was missing a common name.

@kimrutherford I'm not sure if the common name needs to be one word, whether you'd like it quoted, or whether it even needs to be there. Feel free to make any necessary changes.


kimrutherford:
Thanks!

ValWood:
Common name might be useful to keep for the curator maintaining the file,   but might be better as a comment, or optonal...
It might be desirable for display in some instances (although could  be looked up?)

Issue #1403, opened 2018-04-18
Show host and pathogen genes separately when editing genotypes


kimrutherford:
In pathogen/host mode, when there are host genes, a second gene table containing the host genes should appear on the genotype management/edit page.  This list shouldn't appear until the user selects which host organism they are interested in.

kimrutherford:
I've made some progress on this.  I initially thought it would be a good task for @jseager7 and @barnabynorman.  I had a look at the code and it wasn't great.  While improving it I've added some support for having a separate list of genes for pathogen and host.

I've removed some of: root/curs/genotype_manage.mhtml and its controller, GenotypeManageCtrl
and moved things into two new directives: genotype-genes-panel and genotype-gene-list

kimrutherford:
When I attempted to demonstrate this to James and Barnaby a week ago I hit a bug.  I've now fixed that so it works like the screenshots below.  If there is more than one pathogen gene it will show a dropdown for pathogens too.  At the moment the host drop-down doesn't show hosts that have no genes.

![host_path_genotype_gene_split_1](https://user-images.githubusercontent.com/90474/39741950-df5487b8-52ef-11e8-8038-20a99b80296e.png)
![host_path_genotype_gene_split_2](https://user-images.githubusercontent.com/90474/39741952-e2e538a0-52ef-11e8-82b5-1928ff51a2af.png)

kimrutherford:
> At the moment the host drop-down doesn't show hosts that have no genes.

This can be fixed by using the new web service from #1448.

kimrutherford:
>  At the moment the host drop-down doesn't show hosts that have no genes.

I think this would be a good task for @jseager7 and @barnabynorman to have a go at.  It's relatively small and self contained so it's a good task for getting up to speed on the front-end code.

barnabynorman:
@kimrutherford I have added a service to get the organisms list, but it seems to respond with:
`{pathogen_or_host: "unknown",`
before the species list. Did I miss something in the request / is this an error?

kimrutherford:
> Did I miss something in the request / is this an error?

It will return "unknown" for all the organisms unless Canto is in pathogen-host mode.

You'll need to set `host_organism_taxonids` and add some organisms to the database.  There is now some documentation about that setting:
https://curation.pombase.org/pombe/docs/canto_admin/configuration_file#host_organism_taxonids

If you're getting "unknown" after setting `host_organism_taxonids` - that's a bug.

[...]

PR #1466, opened 2018-05-23
Document how pathogen organisms are determined

jseager7:
(References #1403)

Just adding an explanation of how every organism that isn't explicitly specified as a host is assumed to be a pathogen by Canto's configuration.

@kimrutherford I'm not sure if I'm right about this only happening "when `pathogen_host_mode` is enabled", so please amend the text if there's a more accurate explanation.


kimrutherford:
Thanks!

PR #1484, opened 2018-06-05
Fix form control overflow on modal dialogs

### LINKED ISSUE #1479: pop up  form does not resolve well at larger screen size
I thought something had changed, then I realised I had my font  a point larger than the default.
Maybe it can be improved.

someting changed

jseager7:
(Fixes #1479)

The rule for setting form control width to 100% was being overridden in the main stylesheet by the following selector:

https://github.com/pombase/canto/blob/0f046bc6fd463891fcfba615569e52b0f207bea1/root/static/css/style.css#L1385-L1388

I've changed it so that automatic width is only set on the selector for `.help-block`.

@kimrutherford for your information, the selectors apply to the following pages:

```
.help-block
root/static/ng_templates/allele_edit.html
root/static/ng_templates/annotation_edit.html
root/static/ng_templates/annotation_evidence.html
root/static/ng_templates/extension_relation_edit.html

.form-control
root/static/ng_templates/allele_edit.html
root/static/ng_templates/annotation_edit.html
root/static/ng_templates/annotation_evidence.html
root/static/ng_templates/feature_chooser.html
root/static/ng_templates/genotype_background_edit.html
root/static/ng_templates/term_name_complete.html
```

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1382,8 +1382,11 @@ a.curs-editable-table-field img, a.curs-editable-table-long-field img
   filter: alpha(opacity=50);
 }
 
-.form-control, .help-block {
+help-block {
   width: auto;
+}
+
+.form-control, .help-block {
   display: inline;
 }
 
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1399,6 +1399,11 @@ help-block {
   max-width: 600px;
 }
 
+.curs-feature-chooser select,
+.curs-evidence-directive select {
+  width: auto;
+}
+
 table tr td.curs-table-row-remove {
   border-style: none;
 }


kimrutherford:
The look of the dialog changes even when the zoom is normal so I'll let @ValWood comment before merging the change.

This is the original:

annotation-dialog-old-1

This is how it looks after the fix:

annotation-dialog-new-1

ValWood:
I prefer the "gene" and "evidence" pull down to be narrower. It is then more obviously a pull down.

Can we get rid of the grey box around the gene menu label?

kimrutherford:
> Can we get rid of the grey box around the gene menu label?

I think that was only because I had just clicked on it.  It was just temporary.

jseager7:
The width of the form fields expands to fill the popup because their `width` is all set to 100% in the stylesheet. If you want them to be a narrower size, I could amend the pull request and set `max-width` to 100% instead. That _might_ keep them at their default sizes but still stop them overflowing the popup at narrower screen sizes.

jseager7:
Nevermind, `max-width` didn't work. But having another look at the code, I think the overflowing problem is fundamentally related to the fact that most, if not all, of the `input` elements have their `size` attribute set (the `textarea` elements also have the `cols` attribute set), which means they won't respect the scaling of the popup.

The only way (that I can currently think of) to solve this problem, if we want the form controls to have minimum width _and_ not overflow the edge of the popup, is to add set a `min-width` CSS class on each of the form controls that should be narrower, and set the `width` to 100% for the other form controls.

jseager7:
I've added some more styling rules to allow the feature field and the evidence field to be narrower. The overriding is a bit messy, but at least it applies:

```css
.curs-feature-chooser select,
.curs-evidence-directive select {
  width: auto;
}
```
Here's how it looks in Chrome:

feature_evidence_auto_sizing

@ValWood I hope that looks ok to you, and @kimrutherford I'm hoping that setting `width: auto` shouldn't cause any layout problems elsewhere, since I'm not technically restricting the width of the form field.

kimrutherford:
Hi James.

I tried your CSS change.  Unfortunately if was changing to look of other forms.  I thought I'd try restricting the CSS changes to the edit dialog and setting overflow: hidden.  It's not a perfect fix but it's a edge case.

I've made a separate pull request with this change: https://github.com/pombase/canto/pull/1495

jseager7:
Closing this, since it's been fixed by #1495 instead.

PR #1488, opened 2018-06-07
Add support for host organisms with no genes

jseager7:
(Implements #1403)

This makes the following key changes to the genotype management page:

* Organisms with no specified genes now can be selected in the organism drop-down menus.

* `genotype_gene_list.html` now shows a message, instead of showing the gene list, when the selected organism has no genes.

* Genes are now retrieved from the organism, rather than the other way round.
  * Redundant processing related to genes has been removed.

- - -

@kimrutherford might need to fix a bug related to Firefox where the gene list isn't being displayed before this can be merged.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1456,3 +1456,7 @@ a:not([href]) {
   margin: 6px;
   margin-bottom: 12px;
 }
+
+.curs-genotype-no-gene-notice {
+  margin-top: 10px;
+}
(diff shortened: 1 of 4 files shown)


kimrutherford:
> @kimrutherford might need to fix a bug related to Firefox where the gene list isn't being displayed
> before this can be merged.

After a annoying hour of trying to work this out, it seems the Firefox was using a cached copy of genotype_gene_list.html, but only sometimes.  It's very odd.  Clearing the cache or running the test Canto on a different port fixed the problem.

So it's all OK.  Thanks for the work.  I'll merge the changes and deploy it on the PomBase test Canto for a bit of final testing.

kimrutherford:
I noticed a small bug in non-PHI mode.  Clicking the "Deletion" button to add a deletion allele was failing.

Fixed in: ccca998de7b452390b6269fff2a6df4131fc573e

PR #1495, opened 2018-06-12
Attempt to fix overflowing edit dialog

kimrutherford:
Refs #1479

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1396,6 +1396,14 @@ a.curs-editable-table-field img, a.curs-editable-table-long-field img
   display: inline;
 }
 
+.curs-annotation-edit .curs-edit-wide-field {
+  width: 100%;
+}
+
+.curs-annotation-edit .curs-dialog-body-contents {
+  overflow: hidden;
+}
+
 .modal-dialog {
 /*  margin-top: 5%;*/
 }
--- a/root/static/ng_templates/annotation_edit.html
+++ b/root/static/ng_templates/annotation_edit.html
@@ -9,7 +9,7 @@
   </div>
 
   <div class="modal-body">
-    <div ng-switch="annotationType.category">
+    <div class="curs-dialog-body-contents" ng-switch="annotationType.category">
     <table ng-switch-when="ontology" class="curs-definition-table">
 
       <tr ng-if="annotationType.feature_type === 'gene'"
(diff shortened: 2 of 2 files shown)


root/static/css/style.css
```diff
@@ -1396,6 +1396,14 @@ a.curs-editable-table-field img, a.curs-editable-table-long-field img
   display: inline;
 }
 
+.curs-annotation-edit .curs-edit-wide-field {
+  width: 100%;
+}
+
+.curs-annotation-edit .curs-dialog-body-contents {
+  overflow: hidden;
```
jseager7: Setting `overflow` is a good idea. Only thing I'd suggest is using `auto` instead of `hidden`, since even though it adds a horizontal scrollbar at small screen sizes, it might be better to let users have some way to see the clipped content.

root/static/css/style.css
```diff
@@ -1396,6 +1396,14 @@ a.curs-editable-table-field img, a.curs-editable-table-long-field img
   display: inline;
 }
 
+.curs-annotation-edit .curs-edit-wide-field {
+  width: 100%;
```
jseager7: So I guess the plan is to only enable scaling on the wide fields? That's fine by me, since they're the ones that are most likely to cause overflow problems.

root/static/ng_templates/annotation_edit.html
```diff
@@ -106,15 +106,15 @@ <h4 ng-hide="newlyAdded" class="modal-title">
       <tr>
         <td class="title" rowspan="2">Term suggestion</td>
         <td colspan="2">
-          <input class="form-control"
+          <input class="form-control curs-edit-wide-field"
```
jseager7: Is this template used for all modals of this type? My only concern is there might be other modals (or pop-ups) that don't have these CSS classes applied.

root/static/css/style.css
kimrutherford: auto is a good idea.  I just tried it but the scrollbar doesn't disappear when the page it full size.  I can't see what the problem is.  Do you have any idea?

root/static/ng_templates/annotation_edit.html
kimrutherford: There are other dialogs but I was planning to modify them only if they cause problems.

PR #1511, opened 2018-06-18
Implement organism picker and integrate into existing page

barnabynorman:
Tested and appears to be working as expected


kimrutherford:
That's great.  Thanks.

I've tested it in single organism / pombe mode.

PR #1534, opened 2018-06-27
Fix Vagrant permissions on Canto data directory

jseager7:
The configuration for PHI-Canto changed recently to use ` /var/canto-data/` as the data directory, to keep that particular setting in sync between the Vagrant environment and the testing server.

However, on the Vagrant box, only the root user has permissions to create the data directory at `/var/canto-data/`. Permissions have to be transferred to the `canto` user afterwards (the default SSH user), otherwise the `canto_start` script won't work.

Previously the the setup script wasn't run as a privileged user, because of ownership issues, but it's probably more reliable to do everything as root and then just transfer ownership afterwards.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -2,10 +2,10 @@ require 'yaml'
 settings = YAML.load_file './vagrant_config.yaml'
 
 $canto_setup_script = <<-SCRIPT
-  # Go to the Canto source directory
+  # Go to the Canto source directory.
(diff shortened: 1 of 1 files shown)


jseager7:
I'm going to test that this works with a newly-built Vagrant box before I merge.

jseager7:
The changes work correctly, so merging now.

PR #1559, opened 2018-07-11
Revert Pathogen–Host organism pickers

### LINKED ISSUE #1555: Don't show both Pathogen and Host gene pickers in Genotype Management
Now that we've decided to have separate views for Host genotype creation, Pathogen genotype creation, and Meta-genotype creation, we should probably revert the changes that added a second organism and gene picker to the meta-genotype page.

I think the relevant commits are in #1403, but I'm not whether reverting them is the best option; maybe we should just make the changes in new commits.

jseager7:
(Fixes #1555)

This request isn't ready to merge yet, but I'm putting it here to get feedback on the changes.

Since we decided to split the creation of genotypes and the creation of meta-genotypes, some of the changes we made to the Genotype Management page are redundant &ndash; namely the use of two organism pickers.

The key changes are:

* Replaced all of the filtered lists in the `GenotypeGenesPanelCtrl` data with a generic list called `organisms`.
* Updated the controller templates to match the new variable names.
* Added logic in the `GenotypeGenesPanelCtrl` to filter out host organisms with no genes.

The following still needs to be done:

* Move organism filtering logic (e.g. by Pathogen or Host) to `GenotypeManageCtrl`.
* Add some variable or toggle to control whether Pathogens or Hosts are displayed, so that we can have individual links for Pathogen Genotype Management and Host Genotype Management.

Note that the `removeNoGeneHosts()` function that I've added to `GenotypeGenesPanelCtrl` is private to that controller at the moment (it's not bound to `$scope`), but it might be better to extract it as a utility function if we need to filter out host organisms with no genes elsewhere

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -3650,36 +3650,12 @@ var GenotypeGenesPanelCtrl =
         $scope.app_static_path = CantoGlobals.app_static_path;
 
         $scope.data = {
-          allOrganisms: null,
-          hostOrganisms: [],
-          pathogenOrganisms: [],
-          unknownOrganisms: [], // not host and not pathogen
+          organisms: null,
         };
 
         $scope.getOrganismsFromServer = function() {
           Curs.list('organism').success(function(results) {
-            $scope.data.allOrganisms = results;
-
-            $scope.data.hostOrganisms = [];
-            $scope.data.pathogenOrganisms = [];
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks!

I've had a look the and changes make sense.  It all worked when I tested locally too.

So I'm going to merge the changes in the metagenotype-dev branch.  We can fix any problems after that.

Issue #1558, opened 2018-07-11
Make an organism selection directive


kimrutherford:
I thought this was in a directive, but it isn't.  We should extract it so it can be reused:

```html
 <select name="select-organism" ng-model="data.selectedOrganism"
          ng-options="o as o.full_name for o in organisms">
    <option value="">Select an organism ...</option>
 </select>
```
(from genotype_gene_list.html)

jseager7:
The plan is to extract the picker into a directive, allow the picker to fetch its own list of organisms, filter the list (by host or pathogen, etc) based on some criteria, and to allow containing directives to bind to the picker, primarily so they can read the `selectedOrganism` value and react accordingly.

kimrutherford:
I've just pushed some fixes.  The genotype manage page is now a directive.

Probably best to git pull before you do to much.

jseager7:
Thanks for letting me know. I'm up to date now.

jseager7:
Closing this now, since it's been implemented by #1563 and I've tested that it can integrate with the meta-genotype page. Factoring this out has probably caused other bugs with the genotype page that we'll need to fix, though.

PR #1563, opened 2018-07-18
Factor out the organism selector into a directive

jseager7:
(Implements #1558)

This PR extracts the organism selector from `genotype_gene_list.html` into its own directive `organismSelector` (with controller `organismSelectorCtrl` and template `organism_selector.html`).

The new organism selector is responsible for getting its own list of organisms, and it sends the selected organism back to its parent controller using the callback `organismSelected(organism)`. The selector seems to work on the Genotype Management page, but hasn't been tested on the Meta-genotype Management page yet. It should work as expected as long as the controller for the containing directive implements the `organismSelected` method.

Note that this new selector behaves somewhat differently to the old one: the drop-down isn't hidden when an organism is selected, except when there is only one organism in the list. @kimrutherford I'm not sure if you'll be happy with this, but it was the only way I could handle the display of the 'default' organism name in the case where there was only one organism.

@kimrutherford Also, there are conflicting changes in the **pombase/add-organism-selector** branch, which is why I've chosen to merge into **pombase/metagenotype-dev**.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -3526,6 +3526,7 @@ var organismSelector = function ($http, Curs, toaster, CantoGlobals, CantoConfig
   return {
     scope: {
       selectedOrganism: '=',
+      organismSelected: '&',
       genotypeType: '<'
     },
     restrict: 'E',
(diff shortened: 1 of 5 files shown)


jseager7:
Also, let me know if you want me to tidy up the commits. Some of them would be better off squashed together.

PR #1602, opened 2018-08-22
Filter Genotype Management genotype list by selected organism

jseager7:
The genotype list on the Genotype Management page now only shows the genotypes for the organism that is currently selected (using the `organismSelector` directive).

The main changes were to re-index the genotype model by the taxon ID of the organisms, such that the shape of the data now looks like this:

```
genotypeMap = {
  taxonid: {
    singleAlleleGenotypes: [...]
    multiAlleleGenotypes: [...]
  }
  ...
}
```

Other changes include:

* Passing only the single-allele genotypes into `GenotypeGeneList`

* Adding bindings to pass the selected organism back up to the page controller (so that it can filter the genotypes passed down to the genotype list).

* Storing the result of the test of whether to show the message 'No genotypes in this session. Try adding one.' into a scope variable `showNoGenotypeNotice`, since checking `genotypeMap` is now more complex, and probably doesn't need to checked constantly.

@kimrutherford Some of these changes go beyond what you asked for, so feel free to (ask me to) revert them if you think they're excessive or unnecessary. One thing in particular is that I've kept `scope.data` variables around for `singleAlleleGenotypes` and `multiAlleleGenotypes` in `genotypeManageCtrl`, because I thought it would be a bit verbose to reference their index in the `genotypeMap` directly in the template, for example:

```
data.genotypeMap[data.selectedOrganism.taxonid].singleAlleleGenotypes
```
I think I've made it so the other variables just store references to these indexes, so I'm hoping it won't cause a performance impact, but I'm not sure how it works in practice.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4002,10 +4002,20 @@ var genotypeManageCtrl =
 
     var setGenotypes = function (genotypes) {
       $scope.data.genotypeMap = mapGenotypes(genotypes);
-      var selectedOrganismId = $scope.data.selectedOrganism.taxonid;
-      var currentGenotypes = $scope.data.genotypeMap[selectedOrganismId];
-      $scope.data.singleAlleleGenotypes = currentGenotypes['singleAlleleGenotypes'];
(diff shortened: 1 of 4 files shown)


kimrutherford:
Works for me!

PR #1616, opened 2018-09-17
Hide organism selector in single organism mode

kimrutherford:
I'd like to be able to hide the:

    Organism
    Schizosaccharomyces pombe

from the genotype management page when in single user mode.

I tried `ng-if="multiOrganismMode"` as in this pull request but then the gene list disappears.  Is there an easy fix?

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -3813,6 +3813,13 @@ var GenotypeGeneListCtrl =
 
         $scope.hasDeletionHash = {};
 
+        if (!$scope.multiOrganismMode) {
+          Curs.list('organism')
+            .then(function(organisms) {
+              $scope.organismUpdated(organisms.data[0]);
+            });
+        }
+
         $scope.$watch('genotypes',
                      function() {
                        $scope.makeHasDeletionHash();
--- a/root/static/ng_templates/genotype_gene_list.html
+++ b/root/static/ng_templates/genotype_gene_list.html
@@ -1,5 +1,6 @@
 <div class="curs-genotype-gene-list">
   <organism-selector
+    ng-if="multiOrganismMode"
     label="Organism"
     organism-selected="organismUpdated(organism)"
     genotype-type="genotypeType"


jseager7:
I think solving this needs two easy changes:

The 'Organism' label can be hidden by setting the `hideLabel` binding to true:

https://github.com/pombase/canto/blob/39a69b88134a08053304de4be78111539b100672/root/static/ng_templates/organism_selector.html#L6

So the outer template can be amended as follows:

```html
<div class="curs-genotype-gene-list">
  <organism-selector
    hide-label="true"
    organism-selected="organismUpdated(organism)"
    genotype-type="genotypeType"
```

The organism name can be hidden by amending the logic in the template, which is slightly more complicated. Currently the `organism-selector` template just renders out the organism name when there is only one organism:

https://github.com/pombase/canto/blob/39a69b88134a08053304de4be78111539b100672/root/static/ng_templates/organism_selector.html#L14

- - -

This could be changed so that the template only renders the organism name in multi-organism mode, but because `organismSelector` doesn't have `multiOrganismMode` in scope, we might need another binding like `hideOrganismName` that is set according to `multiOrganismMode`. Then the template could be amended to something like:

```
ng-if="(! data.hideOrganismName) && (data.organisms.length === 1)" 
```

I've pushed a commit with these changes to my fork (https://github.com/jseager7/canto/commit/7d9b0cad802054ac5ed2ed041aa7f46d1733352d) since I'm not sure how to properly set up Canto for single organism mode so I can test the changes. Let me know if you'd rather I open a different pull request or if I should just go ahead and commit to `metagenotype-dev`.

kimrutherford:
I tried short circuiting things by setting the organism in GenotypeGeneListCtrl: bb962d5d195a58cd

Can you see any problem with that plan?  It saves adding the organism-selector when it's not needed.

jseager7:
I think your solution is better for single organism mode, since there should be no way for a user to ever add another organism so we don't need to keep the `organismSelector` around in case of that. I've checked the code for `GenotypeGeneListCtrl` and `organismSelector` and I can't foresee any problems. Thanks.

kimrutherford:
Thanks James.  I'll merge this and then we can do proper testing.

Issue #1614, opened 2018-09-11
Wrong screen after deletion


barnabynorman:
When I choose to use the host genotype link that takes me to: host_genotype_manage page and I select a host and make a deletion, it drops me into the old genotype_manage page, so I don't see the results of my deletion.

jseager7:
Closing now, since this was fixed by commit bdbaa5a in pull request #1629.

PR #1629, opened 2018-10-11
[WIP] Fix incorrect links to genotype_manage

### LINKED ISSUE #1606: window.location.href doesn't match pathogen/host genotype pages
(Possibly related to #1604)

Some of the AngularJS code manually sets the `window.location.href` to a hyperlink containing the directory `genotype_manage`, but this isn't used for PHI-Canto anymore since we split the genotype management pages: the correct paths should include `host_genotype_manage` and `pathogen_genotype_manage` instead.

The genotype management page is reset to the 'default' mode any time the `window.location.href` is changed to point at `genotype_manage`, which effectively reloads the page, resets the organism selector, and all hosts and pathogens are shown in the selector from then on.

Here's an example of the faulty code:

https://github.com/pombase/canto/blob/84792ec8f1e6b53f7e6c4a00f4eddecfdeb7ce4c/root/static/js/canto-modules.js#L3585-L3588

I guess there will need to be extra logic to detect what mode the genotype page is in before the `window.location.href` is set &ndash; we possibly _don't_ need to directly check whether we're supposed to be in Pathogen&ndash;Host mode, since we can infer that when the URL contains  `host_genotype_manage` or `pathogen_genotype_manage`. Alternatively, maybe we could just copy more of the current URL (including the genotype manage directory) and simply append the fragment (e.g. `/select/` or `/edit/`)?

jseager7:
(Will fix #1606 once finished)

This pull request replaces references to `genotype_manage` with the correct path for pathogen&ndash;host mode (either `host_genotype_manage` or `pathogen_genotype_manage`).

So far, I've fixed bugs caused by the following actions:

- Creating an allele (#1614)
- _Edit_, on a single or multi-allele genotype
- _Copy and edit_, on a single or multi-allele genotype

Other actions, such as viewing annotations and starting a phenotype annotation, still contain incorrect links to `genotype_manage`. The problems with these are covered [here](https://github.com/pombase/canto/issues/1606#issuecomment-428924623).

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4571,6 +4571,22 @@ var genotypeListViewCtrl =
       replace: true,
       templateUrl: app_static_path + 'ng_templates/genotype_list_view.html',
       controller: function($scope) {
+
+        function getOrganismType(genotypes) {
+          if (CantoGlobals.pathogen_host_mode === "1") {
+            var genotype = genotypes[0];
+            if ('organism' in genotype) {
+              var organism = genotype.organism;
+              if ('pathogen_or_host' in organism) {
+                return organism.pathogen_or_host;
+              }
+            }
+          }
+          return 'normal';
+        }
+
+        $scope.organismType = getOrganismType($scope.genotypeList);
+
         $scope.checkBoxChecked = {};
 
         $scope.columnsToHide = { background: true,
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks James.  It looks good to me so I'll merge your changes.

PR #1642, opened 2018-10-18
Rename disease_formation_phenotype to new name

### LINKED ISSUE #1641: Rename 'disease_formation_phenotype' namespace to match PHIPO
The latest version of PHIPO has renamed the 'disease_formation_phenotype' namespace to 'pathogen_host_interaction_phenotype'. The PHI-Canto configuration files will need to be updated to match.

@kimrutherford I realise that 'pathogen_host_interaction_phenotype' is a particularly long property name. For annotation types, does Canto reference the 'name' property, or the 'namespace' property? If it references `name`, we could abbreviate it to `phi_phenotype` for the sake of the code, and set `namespace` to the full 'pathogen_host_interaction_phenotype'.

@barnabynorman If you're thinking of automatically loading the PHIPO ontology, the latest version will likely break meta-genotype annotations until this issue is fixed. In the short term, you can load an older version of the OBO file from here:

https://github.com/PHI-base/phipo/blob/b0fdc09ba80b266feb4682f203d396990614df0c/phipo.obo

jseager7:
(Fixes #1641)

The latest version of PHIPO has renamed 'disease_formation_phenotype' to 'pathogen_host_interaction_phenotype'. Unfortunately, 'disease_formation_phenotype' is hard-coded in the meta-genotype template and controller.

For now, I've updated the namespace names so that they match the new PHIPO version.

@kimrutherford is there a way we could avoid hard-coding this name? Could we retrieve the correct annotation type from the `CantoConfig` service, e.g. by looking for annotations with `feature_type: 'metagenotype'`?

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -6657,7 +6657,7 @@ var metagenotypeListRow = function(CantoGlobals, Metagenotype) {
 
             $scope.read_only_curs = CantoGlobals.read_only_curs;
             $scope.createAnnotationUri = CantoGlobals.curs_root_uri + '/feature/metagenotype/annotate/' + $scope.metagenotype.metagenotype_id
-                + '/start/disease_formation_phenotype/';
+                + '/start/pathogen_host_interaction_phenotype/';
             $scope.viewAnnotationUri = CantoGlobals.curs_root_uri + '/feature/metagenotype/view/' + $scope.metagenotype.metagenotype_id;
             if (CantoGlobals.read_only_curs) {
                 $scope.viewAnnotationUri += '/ro';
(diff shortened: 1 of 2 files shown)


kimrutherford:
> is there a way we could avoid hard-coding this name? Could we retrieve the correct annotation type
> from the CantoConfig service, e.g. by looking for annotations with feature_type: 'metagenotype'?

Yep, we should be able to do that.  I'll have a look at it tomorrow.

We'll need to handle the potential case where there is more than one ontology/namespace for metagenotypes.

kimrutherford:
I think your change is the right short-term fix.  I've created a separate issue about a longer-term fix: #1644

Issue #1485, opened 2018-06-05
Add separate genotype sections on curation summary page


jseager7:
(Related to #1477)

Following a discussion with @kimrutherford about where the _Multiple-allele_ links are located, we feel it would make more sense to have different sections on the curation summary page for gene-focused workflows and genotype-focused workflows.

The proposal is to add another heading for 'Add genotypes' (or 'Add or view genotypes') and then list genotype-related links under that heading: for example, a link to the genotype management page, and in the case of PHI-Canto, links to lists of host genotypes, pathogen genotypes, and pathogen–host genotypes. Here's a mockup:

![gene_and_genotype_sections](https://user-images.githubusercontent.com/37659591/40975620-70aec4a6-68c3-11e8-98ea-a50c68615ecf.png)

I think we might need to make the _Genotype Management_ link even more obvious than it is now, perhaps by styling it like a button.

If this change was implemented, we could then remove the _Multiple allele_ links from the _Choose curation type_ page, and make it so that when you click a gene, all you see are the gene-focused annotation links. If this is likely to confuse curators who used the _Multiple allele_ links, we could add a link for convenience below the gene annotation links, like so:

![genotype_management_gene_page_link](https://user-images.githubusercontent.com/37659591/40975967-880b4e70-68c4-11e8-9452-5f82fbb5ef8b.png)

- - -

@ValWood please let me know what you think, in particular if you think any of the text is inaccurate or misleading.

kimrutherford:
I suggest these headings under the "Annotate genotypes" heading:

  - Host genotypes and phenotypes ...
  - Pathogen genotype and phenotype ...
  - Pathogen-host interactions and phenotype ..

Each link could go to the genotype management page, but with the management page in the appropriate mode (host genotype mode, pathogen genotype mode or pathogen+host).

ValWood:
I like the  proposal to have a heading for  "annotate genes" and one for "annotate genotypes"

 I think annotate genotypes should offer the 3 options that Kim suggests and provide the appropriate mode.

jseager7:
I've opened a pull request that implements the first part of this issue. Here is how the summary page will look once it's merged:

![summary_genotype_heading](https://user-images.githubusercontent.com/37659591/47495253-904e5680-d84b-11e8-8cd7-2402fb89edd2.PNG)

jseager7:
The changes are merged now, so I'm closing this issue. The genotype heading doesn't have any help text at the moment, so I've opened a new issue about that here: #1658.

PR #1649, opened 2018-10-25
Add genotype section heading to summary page

jseager7:
(See #1485)

This adds an extra heading on the curation summary page to separate the genotype links from the gene links. Here's how it looks:

summary_genotype_heading

I've also ensured that the genotype link supports read-only mode, in the same way that the gene link does:

genotype_heading_read_only

Unfortunately the genotype heading is missing help text, because I don't know how to configure the text. It's possible that this heading doesn't need it though, since the genotype links are more self-explanatory.

@kimrutherford I've copied your `curs-box-title` pattern from elsewhere in the template. I checked `style.css` to make sure there weren't any contextual selectors affecting how `curs-box-title` was styled (e.g. whether it depended on being nested within a `curs-box`), and I couldn't see any. Please let me know if there's anything wrong with the markup.

**Note**: I've used `metagenotype-dev` as the base branch, but if you'd prefer this to be merged to `master`, I can rebase my commits.

### DIFF
--- a/root/curs/front_gene_section.mhtml
+++ b/root/curs/front_gene_section.mhtml
@@ -52,7 +52,11 @@ Edit gene list ...
 % if ($genotype_annotation_configured) {
   <div class="curs-box-title">
     <div>
-      Annotate genotypes
+% if ($read_only_curs) {
+Genotypes from this publication
+% } else {
+Annotate genotypes
+% }
     </div>
   </div>
 %   if ($pathogen_host_mode) {
--- a/root/curs/front_gene_section.mhtml
+++ b/root/curs/front_gene_section.mhtml
@@ -39,7 +39,7 @@ Annotate genes
 % }
   <div style="margin-top: 20px;">
 % if (!$read_only_curs) {
-  <div class="feature-list-action">
+  <div class="feature-list-action" style="margin-bottom: 20px;">
     <a href="<% $edit_path %>">
 %   if ($multi_organism_mode) {
 Edit genes and organisms ...
--- a/root/curs/front_gene_section.mhtml
+++ b/root/curs/front_gene_section.mhtml
@@ -9,13 +9,11 @@ var organismsAndGenes = <% $organisms_and_genes_js |n %>;
 <div class="curs-box">
   <div class="curs-box-title">
     <div>
-% my $and_genotypes = '';
-% $and_genotypes = ' and genotypes' if $genotype_annotation_configured;
 
 % if ($read_only_curs) {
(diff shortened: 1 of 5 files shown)


jseager7:
Looks like Travis-CI is failing because the heading text is linked to some tests. @kimrutherford are the failing tests only checking that the headings have rendered on the curation summary page? The only reference I can find to the text 'Annotate genes' is on the `front_gene_section.mhtml` template.

I'm happy to try to fix the tests if that's the case.

jseager7:
> Could you update the tests? This line in t/90_curs_annotation_check.t needs changing:

Sure, but there are similar references in all the following tests. Should I update all of them?

https://github.com/pombase/canto/blob/3dc42e377dc0b068bfc021c5920ea18569a9c7ec/t/90_curs_controller_actions.t#L253

https://github.com/pombase/canto/blob/3dc42e377dc0b068bfc021c5920ea18569a9c7ec/t/90_curs_pause_annotation.t#L46

https://github.com/pombase/canto/blob/3dc42e377dc0b068bfc021c5920ea18569a9c7ec/t/90_curs_pause_annotation.t#L140

https://github.com/pombase/canto/blob/3dc42e377dc0b068bfc021c5920ea18569a9c7ec/t/90_curs_finished_annotation.t#L44

kimrutherford:
> Sure, but there are similar references in all the following tests. Should I update all of them?

I didn't see those.  Yep, they need fixing.

Since I was looking at them, I had a go at fixing the failures: https://github.com/jseager7/canto/pull/1

jseager7:
@kimrutherford I've merged in your changes, and I've added checks for the genotype management heading in both _edit_ and _read-only_ modes. The read-only checks are in [`t/90_curs_finished_annotation.t`](https://github.com/pombase/canto/pull/1649/commits/a5e7e2f9bcfdd55d376d2a015dc600076d2bc5c1#diff-5f397129ed22d41328c8d21a63629868).

Unfortunately the tests are still failing because of this:
```
Parse errors: Bad plan.  You planned 27 tests but ran 29.
```
I presume I can fix this by changing the number in the `use Test::More tests` statement &ndash; should I increment the value by 1 in each of the failing tests? Should that be done any time a test is changed? I'm not sure I understand the purpose of the statement.

kimrutherford:
> I presume I can fix this by changing the number in the use Test::More tests statement

Yep.  If you add extra tests the number of expected tests needs to be changed.

So in this case change the expected number (after "use Test::More") to 29:

```
Parse errors: Bad plan.  You planned 27 tests but ran 29.
```

PR #1663, opened 2018-11-06
Remove phenotype annotation links from gene page

### LINKED ISSUE #1477: Do we still need mutiple-allele annotation links?
(Possibly related to #1456)

I couldn't see any existing issues that specifically discuss this, but @ValWood mentioned [in issue #1463](https://github.com/pombase/canto/issues/1463#issuecomment-391004436) that we might not need links for _multiple allele_ phenotype annotation types anymore, presumably since they all link to the _Genotype Management_ page.

This is more of a problem for the PHI-Canto version, since there are three different phenotype annotation types, and this means six links for single and _multiple allele_ modes:

phipo_annotation_list

It might be a bit more confusing for the curators to simply remove the links; maybe for PHI-Canto we should have one 'multi-allele phenotype' link that goes to the _Genotype Management_ page, and keep the single allele annotation types specific to the PHIPO namespaces?

Here's a mock-up of what I mean:

phi-canto_annotation_types

jseager7:
(Fixes #1477)

The phenotype annotation shortcuts on the gene page weren't working well in PHI-Canto, since both the single- and multiple-allele phenotype links were repeated for each of the three PHIPO namespaces (making six links in total). Also, the multiple-allele links had the old `genotype_manage` page as their destination, which is related to the problem I was trying to solve in #1606.

For now, we decided to simply remove the links, which I've done by removing the template code. (@kimrutherford if there's a better way to handle this through the Canto configuration, please let me know.)

This change will only affect PHI-Canto for now, since we're merging to `metagenotype-dev`, but we might want a better fix in place before merging that branch to `master`. In terms of what this better fix could be, I've included some thoughts in #1477 (see comment).

### DIFF
--- a/root/curs/gene_page.mhtml
+++ b/root/curs/gene_page.mhtml
@@ -57,19 +57,6 @@ Choose curation type for <% $gene->display_name() %>:
 %     }
     </li>
 %    }
-%   } else {
-  <li class="annotation-type">
-    <a href="#" ng-click="singleAlleleQuick('<% $gene->display_name() |n %>', '<% $gene->primary_identifier() |n %>', <% $gene->gene_id() |n%>, '<%$annotation_type_name%>', <% $taxon_id %>)">
-      Single allele <% ucfirst $type_display_name %>
-    </a>
-    <help-icon key="gene_page_single_allele"></help-icon>
-  </li>
-  <li class="annotation-type">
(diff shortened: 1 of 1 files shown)


kimrutherford:
> which I've done by removing the template code

That looks right to me.

PR #1666, opened 2018-11-07
Set width on organism selector to prevent overflow

jseager7:
The organism names retrieved from UniProt sometimes include all their strains in the name, which makes the organism selector long enough to run into the next column. This pull request fixes that by setting `width: 100%` on the `select` element, ensuring the selector should always have a maximum width that matches its nearest container.

#### Before

orgselector_overflow

#### After

orgselector_no_overflow

### Metagenotype management

This page also uses the organism selector, and the fix is maybe a bit less graceful here because the columns are wider. It still prevents overflow though, and makes the columns more visually consistent. I might be able to set `max-width` instead, if we don't want the `select` to get too big.

orgselector_metagenotype

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1507,3 +1507,7 @@ a:not([href]) {
 .gene-button-delete-text-disabled {
   color: whitesmoke;
 }
+
+.curs-organism-selector-select {
+  width: 100%;
+}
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks James.  Hopefully this will be less of a problem once we remove the strain text from organism names.  But it's good to have a backstop.

PR #1674, opened 2018-11-13
Add success notification when storing annotation

### LINKED ISSUE #1665: No feedback when copying annotation to different genotype
I've noticed that when using 'Copy and edit' on a phenotype annotation (done from the `genotype_page.mhtml` template), and copying the annotation to a different genotype (e.g. changing the genotype from TRI5Δ to TRI5+), there is little to no feedback about the results of the action. You get a toaster notification about the annotation being stored, but there is no confirmation of success.

This is important in this case, because the new annotation can't appear on the current page: the page can only show annotations for the original genotype, not the genotype that the annotation is copied to. What currently happens is that the modal disappears and the page remains unchanged.

I can think of two obvious solutions:

* Redirect the user to the correct genotype page after the annotation is created, or

* Display another toaster message that confirms the annotation was copied, and if possible, provide a link in the toaster box to view the annotation on the correct genotype page.

I think the second option would probably be less confusing for the user, since we can't guarantee that the user wants to be redirected to a new page, and it's not obvious that a redirect should happen from the workflow.

## Relevant code

The directive that handles the annotation table is `annotationTableRow`, and its template is dynamically set based on an `annotationTypeName` attribute:

https://github.com/pombase/canto/blob/c4a3aed8093f7d5d544cf5ad399d208d02e92508/root/static/js/canto-modules.js#L5621-L5624

In this case, the template is `annotation_table_ontology_row`, which calls the `duplicate()` method:

https://github.com/pombase/canto/blob/c4a3aed8093f7d5d544cf5ad399d208d02e92508/root/static/ng_templates/annotation_table_ontology_row.html#L109-L110

This creates a modal controlled by `AnnotationEditDialogCtrl`, and that controller is responsible for creating `toaster` notifications:

https://github.com/pombase/canto/blob/c4a3aed8093f7d5d544cf5ad399d208d02e92508/root/static/js/canto-modules.js#L5194-L5199

@kimrutherford  Could I add a second toaster notification in the case this storing promise succeeds?
```js
q.then(function(annotation) {
  $uibModalInstance.close(annotation);
  // add a toaster notification here
})
```

jseager7:
Fixes #1665 

The 'Copy and edit' link on a phenotype annotation provided no feedback when copying the annotation to a different genotype (e.g. changing the genotype from TRI5Δ to TRI5+). I've fixed this by adding a toaster notification in the success case that notifies the user that the 'Annotation stored successfully.'

We considered adding a link to the relevant genotype details page, so that the user can confirm that their genotype was copied with the expected details, but we decided to put this on hold for the following reasons:

* users may not notice the link in the notification;

* the code I've changed applies to every case where the annotation is stored, and restricting the notification to the case where the genotype is duplicated to a different 'page' is more complicated; and

* the copying of an existing annotation onto a new genotype might be a fairly rare use case; we haven't had any complaints about this lack of functionality yet, so it may not be worth further effort.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -5199,6 +5199,12 @@ var annotationEditDialogCtrl =
       });
       q.then(function(annotation) {
         $uibModalInstance.close(annotation);
+        toaster.pop({
+          type: 'success',
+          title: 'Annotation stored successfully.',
+          timeout: 5000,
+          showCloseButton: true
+        });
       })
       .catch(function(message) {
         toaster.pop('error', message);


kimrutherford:
Thanks!

Issue #1661, opened 2018-11-06
Add genotype creation shortcuts to metagenotype page


jseager7:
Since we changed the metagenotype page's back button destination to the summary page (#1646), it's more inconvenient for users to create a new genotype from the metagenotype page &ndash; they have to go back to the summary and then into one of the genotype management pages.

I'm considering adding shortcut links in each of the genotype picker columns (host and pathogen) that point to the corresponding genotype management page. It would look like this:

![mg_genotype_shortcut](https://user-images.githubusercontent.com/37659591/48063421-e3fe5f80-e1bc-11e8-9757-8f282319c6af.png)

- - -

@kimrutherford My current plan is just to have these link back to the genotype management page, but I thought it might benefit users if the link passed a reference to the organism that was selected on the _metagenotype management_ organism selector, then automatically selected the same organism in the organism selector on the _genotype management_ page.

However, I'm not sure what Canto's convention is for passing this sort of state data around. Is there a webservice I can use to store this state, or could I pass it as a URL parameter?

kimrutherford:
> I'm considering adding shortcut links in each of the genotype picker columns

I think that's a good plan.

> Is there a webservice I can use to store this state, or could I pass it as a URL parameter?

I think it would need to be passed in the URL because it's a different AngularJS page.

jseager7:
Closing this now, and opening new issue (#1713) for future improvements.

PR #1698, opened 2018-11-27
Add genotype management shortcut to metagenotype page

jseager7:
(Partially implements #1661)

This adds links to the genotype picker columns on the metagenotype page to allow users to more quickly navigate back to the pathogen or host genotype management pages. I haven't yet implemented the feature of the genotype management page automatically selecting the organism that was selected on the metagenotype page, since we have to pass this information via the URL and I was concerned that it could conflict with the existing URL logic that handles visually highlighting the last created genotype, for example:

`http://localhost:5000/curs/5b38aa5238216ac8/pathogen_genotype_manage#/select/1`

@kimrutherford I chose to inject `CantoGlobals` to get access to `curs_root_uri`, but it seems that `curs_root_uri` is already on the window object, so I could just access it directly &ndash; this direct access doesn't seem like a good idea to me though. Let me know if you think otherwise.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1511,3 +1511,7 @@ a:not([href]) {
 .curs-organism-selector-select {
   width: 100%;
 }
+
+.metagenotype-genotype-shortcut {
+  padding-top: 20px;
+}
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -6594,7 +6594,14 @@ var metagenotypeGenotypePicker =
           typeLabel: 'Host',
           genotypeType: 'host',
           isHost: false,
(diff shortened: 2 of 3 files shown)


jseager7:
Note that the implementation is also quite brittle; I haven't added any checks for what should happen when `data.genotypeType` or `data.genotypeShortcutUrl` are null, for example. Let me know if you'd prefer some more checks to guard against the link breaking.

kimrutherford:
> this direct access doesn't seem like a good idea to me though. Let me know if you think otherwise

Yep, I think CantoGlobals is the way to go.  It keeps all the global variable access in one place.

kimrutherford:
Thanks.  The change looks good to me.

Issue #1687, opened 2018-11-14
Improve display of strain selector in gene / organism table


ValWood:
This is slightly confusing

![strain selection](https://user-images.githubusercontent.com/7359272/48495764-9622d080-e828-11e8-90ef-69e9b7c36401.jpg)

I think a simple improvement would be to make the strain selection and the genes in  in cleary separate table rows.

jseager7:
If I've understood correctly, the gene and organism table would look something like this:

![confirmation_strain_row](https://user-images.githubusercontent.com/37659591/48551949-e2c0e700-e8ce-11e8-9971-7b729926e841.PNG)

That's just a mockup, so I didn't bother matching the colours of the existing gene table, but if you prefer the colours of the table above, we could change the existing table to match.

jseager7:
I've fixed this on my fork now, here's how it looks:

![1676-compact-table](https://user-images.githubusercontent.com/37659591/51177783-840cef00-18b8-11e9-98d0-a47c6be0fc2d.PNG)

I'd like to include this change as part of a pull request that also fixes #1676, see [my comment](https://github.com/pombase/canto/issues/1676#issuecomment-454363747) in that issue for more details. However if it looks like it's going to take a long time to fix issue 1676, I can implement this separately.

ValWood:
This looks  much better...

jseager7:
Closing now; fixed by the pull request above.

PR #1730, opened 2019-01-15
Redesign the organism table for compactness and consistency

### LINKED ISSUE #1676: Compact the gene and organism list tables
Could you make the "genes" tables slightly more compact?

I only added  a single pathogen and a single host gene, and my screen is already full:

compact this

jseager7:
(Fixes #1676 and #1687)

Val Wood made a request for the organism table (on the 'Gene and host organism list' page) to take up less screen space, and also requested the strain selector to be more clearly separated from the rows containing the organisms's genes. I've done both of these things, and I've also switched the organism table to use consistent styling with the rest of Canto's tables. Here's how the organism table looks now:

1676-row-striping-done

Compared to how the table looked before:

1676-old-organism-table

Note that I've added a slightly different implementation of row striping, such that the colours alternate by organism rather than by gene (that is, by `<tbody>` rather than `<tr>`). This meant adding a new style class, `organism-table`, that inherits most rules from the `table.list` style class. I've since realised that I should probably have used the `curs-genotype-list` style class instead, since this does almost everything I need already &ndash; the only problem is that the row striping is in the opposite sequence to what I have now (i.e. odd rows are coloured, not even), and I think having even rows coloured is clearer in this case. @kimrutherford would you prefer if I re-used the existing `curs-genotype-list` class?

- - -

Note that I've also made some slight changes to the design of the text that shows when an organism has no genes. Here's how it looked before:

1676-no-gene-before

And here's how it looks now:

1676-no-gene-after

My reason for the above change is that I thought it would make row groups with zero genes very visually distinct from rows with one gene. Previously I wasn't sure if the rows with zero genes stood out enough.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -371,6 +371,10 @@ table.curs-genotype-list tbody.odd, table.curs-genotype-list tbody:nth-child(odd
 .organism-table > tbody:nth-child(odd) {
   background-color: #EEEEDD;
 }
+.organism-table__no-gene-notice {
+  text-align: center;
+  color: #555;
+}
 .list_page th {
   white-space: normal !important;
 }
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -331,7 +331,9 @@ div#footer .sponsor, .canto-front-page .footer .sponsor {
   padding: 10px;
 }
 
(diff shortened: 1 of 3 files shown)


ValWood:
The compaction is good...much clearer too

PR #1734, opened 2019-01-17
Allow editing strain of genotype, and refactor strain selector

### LINKED ISSUE #1701: Allow editing strains when editing genotype
Currently there's no way to edit the strain of a genotype after it's been created. We need to add a field to display and edit the strain of the organism in the template [`genotype_edit.html`](https://github.com/pombase/canto/blob/cb4d0870c098d7e2a6a679681d8006b2ac624c49/root/static/ng_templates/genotype_edit.html), and add corresponding logic to the [`genotypeEdit`](https://github.com/pombase/canto/blob/53b9c6d6a0b8fd0aacadb7845bbec57a8a748a59/root/static/js/canto-modules.js#L3253) directive.

There's two possible places the strain picker could be located:

**Option 1:** Insert the picker below the other fields that apply to the whole genotype:

edit_strain_picker_1

**Option 2:** Insert the picker on the edit dialog for every allele in the genotype:

allele_edit_link
allele_edit_strain_picker

The problem with Option 2 is that it allows users to edit the strain for a single allele in the genotype, which is presumably invalid since it makes that allele inconsistent with the multi-allele genotype as a whole. If a genotype should only ever have one strain, then Option 1 makes more sense.

Note that we will also need to guard against users inadvertently making the genotype identical to an existing strain by changing the strain. There's an issue for this at #1620.

jseager7:
(Fixes #1701)

This pull request allows editing the strain of a genotype on the genotype edit page, and extracts the strain selector into its own directive &ndash; it was previously hard-coded into the allele creation modals on the genotype management page. Here's how the new selector looks:

1701-genotype-edit-strain-selector

@kimrutherford The main problem I had implementing this was getting the selector to cooperate with the `StrainsService`. I considered changing the internals of the service, but that risks affecting many more directives than the ones I was focusing on, so eventually I decided to switch the relevant directives back to using the `Curs` service &ndash; specifically, `Curs.list('strain')` &ndash; and performed the filtering of strains (by taxon ID) in the controllers. I only intend for this to be a workaround until we can figure out a more permanent fix to the `StrainsService`. Please let me know if you want further explanation / justification.

I've also added the ability for the strain selector to be styled according to its context, by optionally passing in a style class name to the directive (as a text binding). I thought this was necessary since Canto's modals use the bootstrap-style `form-control` classes, whereas the fields on the _genotype edit_ page use minimal styling. I've set `form-control` as the default, and I've specified other styling rules where necessary, so the selector always looks consistent with its context. (It's possible that I could've used CSS descendant selectors to achieve the same result, but this seemed more complicated at the time.)

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1536,6 +1536,10 @@ a:not([href]) {
   display: inline-block;
 }
 
+.strain-selector__select--genotype-edit {
+  padding: 3.5px 0;
+}
+
 .metagenotype-genotype-shortcut {
   padding-top: 20px;
 }
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1532,6 +1532,10 @@ a:not([href]) {
   width: 100%;
 }
 
+.strain-selector__container {
+  display: inline-block;
+}
+
 .metagenotype-genotype-shortcut {
   padding-top: 20px;
 }
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -3005,6 +3005,10 @@ var alleleEditDialogCtrl =
       getStrains: StrainsService.getSessionStrains,
     };
 
+    $scope.strainSelected = function (strain) {
+      $scope.strainData.selectedStrain = strain;
+    };
+
     $scope.env = {
     };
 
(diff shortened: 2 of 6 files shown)


jseager7:
Just realised I forgot to toggle the strain selector off for single organism mode, but that's fixed now. It doesn't seem to break the genotype storage when a strain isn't specified; the strain just stays the same as it was before.

PR #1754, opened 2019-01-29
Show organism name on genotype edit page

### LINKED ISSUE #1749: Show organism name on genotype edit page
(Requested by @kimrutherford)

The genotype edit view now allows the strain of the genotype to be edited, but it doesn't show the name of the organism corresponding to those strains anywhere on the page. We could add the name of the organism to the heading of the page, like so:

1748-genotype-organism-heading

Or we could put the organism name near to the other input fields, so it looks consistent with the other editable fields, but isn't editable itself: 

1748-genotype-organism-field

jseager7:
(Fixes #1749)

This changes the Genotype Edit page to display the scientific name of the organism whose genotype is being edited. Based on feedback from other collaborators, it was decided to show the organism name above the strain picker. The organism name only displays in multi-organism mode.

1749-organism-name

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -736,6 +736,10 @@ table td.curs-genotype-checkbox-column {
   color: grey;
 }
 
+.curs-genotype-edit__organism-label {
+  padding: 3px 0;
+}
+
 .curs-genotype-list .selected, .curs-genotype-details .selected {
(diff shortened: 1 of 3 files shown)


kimrutherford:
Thanks James.  That works for me.

Issue #1760, opened 2019-02-05
Add copy-editing of metagenotype annotations


jseager7:
(Requested by @CuzickA)

Currently, there is no way to change the target metagenotype of a metagenotype annotation (_pathogen&ndash;host interaction phenotype_). Instead, a new annotation has to be created each time, which is a problem for experiments where many interactions display the same phenotypes.

This is how the annotation editing popup looks now:

<img width="573" alt="mg-ant-no-selector" src="https://user-images.githubusercontent.com/37659591/52274267-0af73980-2944-11e9-889b-29b059ca6c82.PNG">

Ideally, there should be a selector listing the metagenotypes added to the session. The organism name may have to be displayed in this list as well, in the case that different organisms have the same gene names. We can either re-use the existing text display for metagenotypes (as seen in the popup title in the image above), or we can use something more compact, for example:

> TRI5&Delta; (F. graminearum, PH-1) &mdash;  OPR3&Delta; (S. lycopersicum, Craigella)

Which would look like this in the selector:

<img width="577" alt="mg-ant-add-selector" src="https://user-images.githubusercontent.com/37659591/52275514-ad64ec00-2947-11e9-92b9-25a59d370509.PNG">

- - -

Note that the metagenotype name may have to account for alleles with more complicated formatting &ndash; using standard parentheses to delimit the organism name and strain might be confusing when those parentheses are in the allele names as well, for example with nonsense mutations:

> unnamed(aa123->stop)

We could solve this by using parentheses that are never used in allele names, for example square brackets `[]` or curly brackets `{}`.

jseager7:
### Relevant code:

- Template `annotation_edit.html`
- Controller `annotationEditDialogCtrl`
- Function `startEditing` (opens modal linked to template and controller)
- Directive `featureChooser`
  - presumably this needs the `featureType` binding to have a value of `"metagenotype"`, but the controller doesn't support this option yet; the `Metagenotype` service needs to be injected and logic needs to be added to `function featureChooserControlHelper`.

ValWood:
Yes this is strange In normal Canto you *can* edit the genotype in this pop up with cop/edit.....

kimrutherford:
> Yes this is strange In normal Canto you can edit the genotype in this pop up with cop/edit.....

Changing the genotype still works.  It's changing the meta-genotype of an annotation that needs implementing.

kimrutherford:
[...]

PR #1769, opened 2019-02-07
Allow editing of metagenotype in annotation_edit

kimrutherford:
Refs pombase/canto#1760

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -1305,6 +1305,8 @@ function featureChooserControlHelper($scope, $uibModal, CursGeneList,
     } else {
       Curs.list('metagenotype').then(function (res) {
         $scope.features = res.data;
+      }).catch(function () {
+        toaster.pop('note', "couldn't read the metagenotype list from the server");
       });
     }
   }
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -1284,7 +1284,7 @@ function openSingleGeneAddDialog($uibModal) {
 
 
 function featureChooserControlHelper($scope, $uibModal, CursGeneList,
-  CursGenotypeList, toaster) {
+  CursGenotypeList, Curs, toaster) {
   function getGenesFromServer() {
     CursGeneList.geneList().then(function (results) {
       $scope.features = results;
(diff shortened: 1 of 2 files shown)


root/static/js/canto-modules.js
```diff
@@ -1284,7 +1284,7 @@ function openSingleGeneAddDialog($uibModal) {
 
 
 function featureChooserControlHelper($scope, $uibModal, CursGeneList,
-  CursGenotypeList, toaster) {
+  CursGenotypeList, Curs, toaster) {
```
jseager7: We have a `Metagenotype` service now that just handles metagenotypes:

https://github.com/pombase/canto/blob/72117a8a84b7e7ed3e752075d840611d3bfc9196/root/static/js/canto-modules.js#L471

It's possible you could use that instead, unless you're already aware of it and decided not to use it because you preferred the way `Curs` worked. Alternatively, we could just inject `Curs` and not use `CursGeneList` or `CursGenotypeList`, since this function and its directive might be using most of the data that `Curs` exposes  anyway. It's your choice.

(As an aside, I think `Metagenotype` should probably be renamed to `MetagenotypeService` or `CursMetagenotypeList` to make its function more obvious.)

root/static/js/canto-modules.js
```diff
@@ -1296,11 +1296,17 @@ function featureChooserControlHelper($scope, $uibModal, CursGeneList,
...
+    if ($scope.featureType === 'genotype') {
+      CursGenotypeList.cursGenotypeList().then(function (results) {
+        $scope.features = results;
+      }).catch(function () {
+        toaster.pop('note', "couldn't read the genotype list from the server");
+      });
+    } else {
+      Curs.list('metagenotype').then(function (res) {
+        $scope.features = res.data;
+      });
```
jseager7: I presume there needs to be a catch block after `Curs.list('metagenotype')`, like the call to `cursGenotypeList()` above?

root/static/js/canto-modules.js
kimrutherford: I didn't use the Metagenotype Service because it has quite a different interface to the other two services and it didn't fit the usage.

I like that the Metagenotype Service uses $rootScope.$broadcast() to notify about changes.  I think that's a good plan.  I didn't realise that was possible when I implemented the Curs service.

I think it would be a good clean up task to improve Curs, CursGeneList etc. to use $broadcast() and then make the interface of the Metagenotype Service match the other Services.

kimrutherford: > As an aside, I think Metagenotype should probably be renamed to MetagenotypeService

I'd be happy with having "Service" at the end of all services.  I haven't been very consistent.

root/static/js/canto-modules.js
kimrutherford: Thanks.  I've added a .catch()

PR #1777, opened 2019-02-12
Make quick deletion strain picker conditional

### LINKED ISSUE #1771: Single organism mode: hide quick deletion strain selector
In single organism mode, when you make a quick deletion on the genotype management page, using the 'Deletion' shortcut in the gene list:

single-org-quick-deletion

It still shows the strain selector, except there are no strains, so you can't finish the quick deletion and the button is effectively useless:

single-org-strain-selector

There should be a check here to skip the modal for modes where strains aren't used, or when there are no strains added to the session.

jseager7:
Fixes #1771 

This pull request adds a dispatch function to `GenotypeGeneListCtrl`, that assigns a function to the `ng-click` directive on the quick deletion button (on the Genotype Management page), based on the value of `$scope.multiOrganismMode`.

@kimrutherford let me know if you're not keen on the dynamic function assignment to the new `$scope.quickDeletion` variable, or if you'd prefer to not use a ternary expression. I admit the code is a bit esoteric:

```js
$scope.quickDeletion = $scope.multiOrganismMode
  ? $scope.deleteSelectStrainPicker
  : $scope.makeDeletionAllele
```

I did this in the interest of efficiency, so the function doesn't have to keep checking what mode it's in every time the 'Deletion' button is clicked. If you'd prefer a more conventional function that simply calls the correct function, I can amend it. This solution might be a bit brittle because it depends on the function signature of both methods being the same.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4091,6 +4091,10 @@ var GenotypeGeneListCtrl =
           });
         };
 
+        $scope.quickDeletion = $scope.multiOrganismMode
+          ? $scope.deleteSelectStrainPicker
+          : $scope.makeDeletionAllele
+
         $scope.deletionButtonTitle = function (gene_id) {
           if ($scope.hasDeletionHash[gene_id]) {
             return 'A deletion genotype already exists for this gene';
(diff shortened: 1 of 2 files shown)


kimrutherford:
>  If you'd prefer a more conventional function that simply calls the correct function, I can amend it. 

Your implementation looks clear to me.

> This solution might be a bit brittle because it depends on the function signature of both methods being the same.

Moving to TypeScript will solve that!  :-)

PR #1782, opened 2019-02-21
Hide gene phenotype links in pathogen-host mode

### LINKED ISSUE #1781: PHI-Canto still has too many gene curation types
This is a regression caused by commit 6f16368, which was a temporary fix for issue #1765.

I think we might need to change the way phenotype annotations are generated from Canto's configuration, because even after merging ontology namespaces in PHIPO in an attempt to simplify things, Canto still generates the phenotype links from the annotation names in its configuration, which generates way too many links in PHI-Canto's case:

phi-canto_phenotype_links

Ideally, we should have a link that's something like 'Single species phenotype' or 'Single species phenotype (single allele)', to match PHIPO's term &ndash; I'm now of the opinion that the single allele phenotype link is a useful shortcut &ndash; and then have a separate section for multi-allele genotypes that links to the Genotype Management page.

There's more discussion of this in #1477, and a solution to this issue will probably fix that one too.

jseager7:
Fixes #1781 

This change hides the phenotype annotation shortcuts on `gene_page.html` when Canto is in pathogen-host mode, to buy us more time to think of a better fix. There was no reference to `$pathogen_host_mode` in the template, so I added one by following convention from other templates:

```perl
my $pathogen_host_mode = $st->{pathogen_host_mode};
```

### DIFF
--- a/root/curs/gene_page.mhtml
+++ b/root/curs/gene_page.mhtml
@@ -58,6 +58,7 @@ Choose curation type for <% $gene->display_name() %>:
     </li>
 %    }
 %   } else {
+%     if (!$pathogen_host_mode) {
   <li class="annotation-type">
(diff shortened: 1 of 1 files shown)


kimrutherford:
Thanks.  That looks good to me.

PR #1791, opened 2019-03-06
Resize finish textarea to suit laptop screens

### LINKED ISSUE #1784: Final comments box too tall for narrow screens
When you click the 'submit to curators' button, you click a button to submit to curators.

you are then taken to a page where you can add a note and click another button to 'finish'

on a wide screen laptop the 'finish' button is not visible unless you scroll down.

I fear some people will press the submit button without 'finishing'.

can we:
1. make submitting/finishing a one-step process
2. move the comment textbox somewhere inside main canto
3. check how many sessions have annotation/s that haven't been touched for 3months+? (these people will either have given up - or worst case scenario they thought they submitted the work and it wasn't actually submitted?

jseager7:
(Fixes #1784)

This change reduces the default height of the comments text box on the 'Finish Curation' page, so that the 'Finish' and 'Back' buttons are still visible on shorter screens (like laptops):

1784-short-textarea

I've also set a `min-height` on the `textarea` element to prevent users from collapsing the text area to the point that it's almost invisible, and I've restricted resizing the text area to allow only vertical resizing, since horizontal resizing tends to interfere with usability if the user accidentally forces horizontal scrolling by resizing too far. @kimrutherford If you'd prefer some horizontal resizing, I think I can set a `max-width` instead to prevent the text area flowing off the side of the screen.

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -2064,6 +2064,9 @@ sub finish_form : Chained('top') Args(0)
         type => 'Textarea',
         cols => 80,
         rows => 3,
+        attributes => {
+          class => 'curs-final-comments-field'
+        },
         default => $self->get_metadata($schema, MESSAGE_FOR_CURATORS_KEY) // '',
       },
       map {
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -2060,7 +2060,10 @@ sub finish_form : Chained('top') Args(0)
 
   my @all_elements = (
       {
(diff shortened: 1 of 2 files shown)


kimrutherford:
That's great James.  Thanks for diving into the Perl code to make this change.  Very brave!  :-)

> If you'd prefer some horizontal resizing, I think I can set a max-width instead to prevent the text
> area flowing off the side of the screen.

Let's deploy as-is and see what the curators think.

Cheers.

ValWood:
I'm sure it will be fine. The box was much bigger than required, and the size does not dictate a text limit.
v

PR #1793, opened 2019-03-12
Fix confirmation button on edit organisms page

### LINKED ISSUE #1762: Continue button on 'confirm genes' page doesn't disable correctly
The 'Continue' button in `edit_organisms.html` is supposed to prevent the user from leaving the page until at least one strain has been added for each organism. However, even though the button appears disabled, you can still click the button, and doing so takes you to the summary page regardless of whether you've added strains or not.

The function in question is [`$scope.getContiueUrlDisabled`](https://github.com/pombase/canto/blob/616cc98d77d31f9e2f237138a1c555cb0a1803dc/root/static/js/canto-modules.js#L7431) (sic) in the controller of `editOrganisms`. The bug might be related to the fact the controller is interacting with the `StrainsService`, which has undergone some refactoring recently:

https://github.com/pombase/canto/blob/616cc98d77d31f9e2f237138a1c555cb0a1803dc/root/static/js/canto-modules.js#L7437-L7463

### Steps to reproduce

1. Start a new session
2. Add gene `Q00909`
3. Click 'Continue'

jseager7:
(Fixes #1762)

This fixes a bug with the 'Continue' link on the 'Gene and host organism list' page (template `edit_organisms.html`). The bug was due to the `ng-disabled` directive being applied to a link that was _styled_ like a button (rather than being applied to a real `<button>` element), and links do not support the `disabled` attribute.

I've changed the link to a button so that the disabled state is properly supported – click events do not fire from an element in the disabled state, so checks aren't needed in the controller to prevent redirection. @kimrutherford If you'd prefer the element to stay as a link, I think it should be possible to override the redirection on the link to do nothing if the state of organisms and strains is invalid.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7477,7 +7477,7 @@ var editOrganisms = function () {
       $scope.getPathogens = EditOrganismsSvc.getPathogenOrganisms;
       $scope.getHosts = EditOrganismsSvc.getHostOrganisms;
 
-      $scope.getContiueUrlDisabled = function () {
+      $scope.isContinueUrlDisabled = function () {
         if ($scope.getPathogens().length == 0 &&
           $scope.getHosts().length == 0) {
           return true;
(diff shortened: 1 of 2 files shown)


jseager7:
I don't think this change affects single organism mode, since that uses a different view for the gene confirmation page: the 'Continue' button on that page is actually an `<input>` of type `submit`.

kimrutherford:
Thanks James.  I've tested the change and it looks good to me.

kimrutherford:
> I don't think this change affects single organism mode,

Yep, I agree.  I've used the page in single organism mode after merging your fix and it's still working.  :-)

Issue #1805, opened 2019-03-22
Add a strains_mode option


kimrutherford:
We currently show strain selectors in multi organism mode but it's not always needed.  We should probably add an independent setting for it.

@jseager7 I think you suggested this a while ago.  It's now more urgent because we use multi organism for FlyBase-Canto and worrying about strains doesn't make sense in that case.

![Screenshot from 2019-03-22 16-02-55](https://user-images.githubusercontent.com/90474/54798145-11573000-4cbc-11e9-8517-906d7bae137e.png)

kimrutherford:
I've had a go at this (#1809) but it still needs testing.

kimrutherford:
And remove the strains column from the phenotype/genotype annotation tables.

vmt25:
FlyBase does not need this - could it be excluded fro  FlyCanto's genotype workflow and table?

kimrutherford:
> And remove the strains column from the phenotype/genotype annotation tables.

I've made that change too.

> I've had a go at this (#1809) but it still needs testing.

The changes are on a branch.  I'll do more testing tomorrow and then merge things.

> FlyBase does not need this - could it be excluded from FlyCanto's genotype workflow and table?

You'll be able to hide the strains with the new `strains_mode` option once the changes are merged.

kimrutherford:
This working now and it's merged into the master Git branch.

Vitor, to disable strains in your Canto add this line to your `canto_deploy.yaml`:

```
strains_mode: 0
```

I've updated the flybase-test server with this change. eg.:
https://curation.pombase.org/flybase-test/curs/e974e992ec7e2aad

PR #1809, opened 2019-03-29
Add a strains_mode

kimrutherford:
Hi @jseager7 

I've had a go at implementing a strains_mode (#1805).  I think I've fixed all the places that I can see where strains are required.  When you get a chance could you have a quick test and let me know if you spot bugs or any place I've missed?  If you have genotype with strains in your session but strains mode is disabled, the strain name will still appear in the genotype table.

You'll need to add this to your `canto_deploy.yaml` to activate strains mode:

```
strains_mode: 1
```

There's no rush to merge this so perhaps we should wait until after the conference even once any bugs are fixed.

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -30,8 +30,8 @@ app_version: v1279
 schema_version: 26
 
 ## Set this to create a Canto instance for a particular organism.  If not set,
-## all gene lookups with match genes from any organism (and will return the
-## organism).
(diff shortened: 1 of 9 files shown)


kimrutherford:
Hi James.  I don't think these changes will cause any problems for PHI-Canto.  If you notice any bugs I'll fix them as my top priority.

PR #1813, opened 2019-04-02
Replace 'meta-genotype' with 'metagenotype' everywhere

### LINKED ISSUE #1812: Use consistent name for 'metagenotype'
PHI-Canto currently uses different spellings for the word 'metagenotype': sometimes it is hyphenated as 'meta-genotype', whereas sometimes it is just 'metagenotype'. Given that the closest comparable word, 'metagenomics', typically does not use a hyphen, we've decided to use 'metagenotype' as the canonical spelling.

jseager7:
(Fixes #1812)

Canto was previously using inconsistent naming for the term 'metagenotype'. This pull request should replace all those inconsistent representations with the word 'metagenotype'.


kimrutherford:
Thanks James.  I resolved a conflict with a700e2cdd before merging.

Issue #1818, opened 2019-04-02
Hide selector, show all organisms on genotype management page (optionally)


kimrutherford:
I think @vmt25 mentioned this on Skype. FlyBase would like to show all genotypes on the genotype management page and hide the organism selector.  This would need to be an option as PHI-Canto needs the current functionality.

@jseager7 I won't attempt this until you're happy with the genotype management refactoring you're doing. This isn't an urgent issue.

vmt25:
> I think @vmt25 mentioned this on Skype. FlyBase would like to show all genotypes on the genotype management page and hide the organism selector

That would be nice. Every now and then there are multi-allele genotypes that combine genes/alleles of different species.

(It also has the very minor bonus of requiring one less click!)

kimrutherford:
> Every now and then there are multi-allele genotypes that combine genes/alleles of different species.

I've made a separate issue about that: #1826.

For now I'll add an option to remove the organism selector from the genotype management page. I'll disable the combining alleles from different organisms until we've chatted about #1826.

kimrutherford:
I've now added this option.  So in the flybase-test Canto the organism selector is hidden and it now always shows genes and genotypes from all species on the genotype management page.

eg. https://curation.pombase.org/flybase-test/curs/33dac4236dd1bfdb/genotype_manage

I'm going to leave this issue open while I do more testing in pombe-mode and pathogen-host mode.

vmt25:
Looks great! And seems to allow combining single-allele genotypes from different organisms (see #1826 https://github.com/pombase/canto/issues/1826)

kimrutherford:
> I'm going to leave this issue open while I do more testing in pombe-mode and pathogen-host mode.

It all seem OK so I'll close this now.

Issue #1810, opened 2019-03-31
Hide the expression column for Fly-Canto


kimrutherford:
There is no expression information to show in Fly-Canto so we should add an option to hide the expression column in genotype tables.

kimrutherford:
This change is done but not merged into the master branch yet.  @jseager7 is doing some work on the genotype management page and it's easier to wait for that to be done before merging.

I've added a new configuration option to control the expression. Soon we'll be able to add this to `canto_deploy.yaml` to hide the expression everywhere in the interface:

```
alleles_have_expression: false
```

jseager7:
It looks like we've got multiple issues that can't be merged to `master` because they're blocked by my refactoring. However, the changes to the genotype management page are pretty much finished &ndash; although I might end up with a few merge conflicts when I try to merge my changes now.

I'm not sure how long the metagenotype page refactoring will take, but if necessary I could allow us to merge the changes to the genotype page sooner by re-adding the code for the 'old' organism selector, while keeping the new one around as well, and pointing the metagenotype page at the old organism selector and the genotype page at the new one. The metagenotype views are almost completely isolated from FlyBase anyway, so I could then gradually refactor the metagenotype page to replace the old organism selector with the new one.

kimrutherford:
>> Soon we'll be able to add this to canto_deploy.yaml to hide the expression everywhere in the interface:
> `alleles_have_expression: false`

Sorry, that should be:

`alleles_have_expression: 0`

The change is now in the master Git branch.

kimrutherford:
I've fixed the genotype edit page to obey that new setting.  I had forgotten about it.

PR #1825, opened 2019-04-03
Implement refactored organism selector on Genotype Management page

jseager7:
(Partial fix for #1804)

There are a few issues pending for the Genotype Management page (for FlyBase) that are waiting for my refactoring of the organism selector to be finished (#1810, #1818). However, I don't think it's necessary to wait, since the organism selector already works correctly on the Genotype Management page: the problem is it's caused breaking changes to the Metagenotype Management page. Fortunately, FlyBase doesn't use Metagenotype Management page, and the changes to the organism selector were only really needed to enable a link to the Metagenotype Management page.

So for now, I'm suggesting we merge early by keeping _both_ versions of the organism selector (old and new): `genotypeManage` will use the new selector, and `metagenotypeManage` will use the old selector. This way, we can integrate my changes to stop them blocking us from integrating anything else into `master`.

My plan is to rename the `organismSelectorNew` directive and template to replace the (old) `organismSelector` once I've managed to integrate the new selector into the Metagenotype Management page &ndash; this shouldn't prevent us from making the changes that FlyBase need in the meantime. The changes to the Metagenotype Management can be handled on a new pull request.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4034,6 +4034,7 @@ var genotypeManageCtrl =
           editingGenotype: false,
           editGenotypeId: null,
           multiOrganismMode: false,
+          hostOrganismExists: false,
           showMetagenotypeButton: true,
           showNoGenotypeNotice: true
         };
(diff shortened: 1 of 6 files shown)


jseager7:
Also note that I haven't tested this in multi-organism (FlyBase) mode, since I don't know how to configure my local copy of Canto for it.

kimrutherford:
I did some testing and your changes worked for me.  So I'll go ahead and merge.

PR #1827, opened 2019-04-05
Add split_genotypes_by_organism option

kimrutherford:
I've added a new setting to optionally hide the organism selector on the genotype management page.  For now I've made a branch because I haven't tested the code in all the cases yet.

Sorry, there are quite a few lines changed. I think I haven't made the code worse.  :-)

The genotype management controller still looks the same from the outside (same attributes get passed in).  I had to move some logic from GenotypeGeneListCtrl to genotypeManageCtrl.  We now pass a gene list into GenotypeGeneListCtrl instead of `selectedOrganism`.

Refs pombase/canto#1818

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -79,6 +79,10 @@ max_existing_annotations: 100
 # as a column in the genotype table on the genotype management page
 alleles_have_expression: 1
 
+# if 1, show the organism in a selector on the genotype management pages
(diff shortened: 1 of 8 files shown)


kimrutherford:
> For now I've made a branch because I haven't tested the code in all the cases yet.

I did more testing and now it's merged.

PR #1838, opened 2019-04-24
Refactor the metagenotype management page

### LINKED ISSUE #1837: Metagenotype management breaks when custom strains are used
@CuzickA noticed that when 'Unknown strain used' and a custom strain (from the option _Type a new strain..._) are chosen in the strain selector on the _Gene and host organism list_ page, the wild-type strain selector on the Metagenotype Management page doesn't display any strains at all.

In short, an organism configured like this:

custom_unknown_strain_bug_1

Leads to this:

custom_unknown_strain_bug_2

jseager7:
(Fixes #1837)

This pull request refactors the metagenotype page to allow it to interact properly with other components. The refactoring tries to make the code more consistent with the rest of the application, particularly with regards to the refactoring of the Genotype Management page.

Notable changes include:

- Passing more data as objects (rather than passing object properties), through one-way bindings.

- Reducing nesting of components, to reduce the depth of callback binding chains.

- Removing state management from the `Metagenotype` service and transferring this responsibility to the `MetagenotypeManage` controller (might not be an ideal solution long-term).

- Delegating more control of state and data to the overall page controller `MetagenotypeManage`.

- Numerous fixes to page controls, such as the disabling logic on the metagenotype creation button, and selection of custom and unknown strains in the wild-type genotype strain selector.

These changes shouldn't need testing in any other mode, since the components and pages are only used for pathogen&ndash;host mode. I've done manual testing of the critical functionality on the page, and I didn't notice any bugs, but the page may need some more comprehensive testing just to ensure that nothing subtle is going wrong.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7070,9 +7070,8 @@ var metagenotypeGenotypePicker =
       controller: function ($scope) {
 
         $scope.data = {
-          genotypes: null,
-          singleAllele: [],
-          multiAllele: [],
+          singleAlleleGenotypes: [],
+          multiAlleleGenotypes: [],
           wildType: [],
           typeLabel: 'Host',
           genotypeType: 'host',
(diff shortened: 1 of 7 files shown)


kimrutherford:
Thanks for all those improvements.  I've done some testing and it's all working fine for me so I'll go ahead and merge.

PR #1844, opened 2019-04-30
Remove the old organism selector

jseager7:
I forgot to remove some disused organism selector code in pull request #1838. This fixes that. The implementation for `organismSelectorNew` overwrites `organismSelector`. I've tested this in PHI-base and PomBase configurations, but haven't tested for FlyBase.


kimrutherford:
> but haven't tested for FlyBase.

I don't think the FlyBase version uses the organism selector.

I've tested your changes and it all works for me.

PR #1851, opened 2019-05-07
Use flexbox layout on genotype management page

### LINKED ISSUE #1508: genotype table overlapping with gene list
This only just happened

jseager7:
(Fixes #1508)

Previously we were using Bootstrap columns to lay out the genotype management page, but using fixed widths didn't cooperate with the automatic layout on tables: typically the gene list column would overflow into the genotype list column once the gene names got too long. Using a flexbox layout seems to fix this:

genomgmt-flexbox

The changes introduce a few new CSS classes to explicitly tag the flex container and the columns on the genotype management page:

* `curs-gm-container`
* `curs-gm-gene-col`
* `curs-gm-genotype-col`

`gm` is just shorthand for 'genotype management'. If you'd prefer any of these classes to be named differently, please let me know.

Note I've left the Bootstrap `row` class in the markup, although I'm not sure if we necessarily need it anymore.

- - -

This change still needs testing in Firefox and Internet Explorer 11 &ndash; the latter in particular is the most buggy with regards to flexbox implementation, but it doesn't seem like the known issues should affect us, based on this list: https://github.com/philipwalton/flexbugs

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1559,3 +1559,15 @@ a:not([href]) {
   resize: vertical;
   min-height: 60px;
 }
+
+.curs-gm-container {
+  display: flex;
+  justify-content: flex-start;
+  flex-wrap: wrap;
+}
+.curs-gm-gene-col {
+  margin-left: 3px;
+}
+.curs-gm-genotype-col {
+  margin-left: 30px;
+}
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks James.  That's a big improvement.  I have a small patch in progress.

PR #1884, opened 2019-05-21
Fix strain picker displaying nothing on page load

### LINKED ISSUE #1881: Strain picker defaults to null value
Possibly due to the changes in #1874, the strain picker on the gene confirmation page is now blank, rather than showing a default value:

empty-strain-picker

Looking at the markup, it seems like the `<select>` element is somehow being assigned a null object as a value:

```html
<option value="? object:null ?" selected="selected"></option>
```
No idea why this is happening at the moment; I'll investigate further.

jseager7:
Fixes #1881 

It's possible that there's been some changes in the `ngModel` logic for AngularJS that prevents it selecting a disabled `<option>` element when used with a `<select>`. This was causing the strain picker on the gene confirmation page to default to an empty value, since its default `<option>` is given the `disabled` attribute in the template:

empty-strain-picker

This change removes the `disabled` attribute as a quick fix, and has the controller check that the drop-down's selection isn't a 'special' option that represents an action (e.g. 'Type a strain' or the default descriptive 'Add a strain...' option).

@kimrutherford This fix is brittle, since it relies on values that are hard-coded into the template staying in sync with values hard-coded into the controller. If you don't think that's a problem then I'll leave it as is, but I think a better solution might be to either:

- **a)** store all the option text as (scope) variables in the controller and then bind them to the template, or

- **b)** use the `ng-options` directive to render the list instead, since this can set the `disabled` attribute as part of the expression. I expect this would require reshaping the data in the controller (see https://github.com/pombase/canto/issues/1881#issuecomment-494373625 for more caveats).

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7668,7 +7668,7 @@ var strainPicker = function () {
         newStrain: '',
         strains: null,
         sessionStrains: null,
-        strainSelector: 'Add experimental strains for this organism'
+        strainSelector: ''
       };
 
       CantoService.lookup('strains', [$scope.taxonId]).then(function (data) {
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7675,7 +7675,6 @@ var strainPicker = function () {
         $scope.data.strains = data;
       });
 
-
       $scope.getSessionStrains = function () {
         StrainsService.getSessionStrains($scope.taxonId)
           .then(function (sessionStrains) {
@@ -7686,7 +7685,7 @@ var strainPicker = function () {
       $scope.getSessionStrains();
 
       $scope.changed = function () {
-        if ($scope.data.strainSelector !== 'Type a new strain') {
+        if ($scope.data.strainSelector && notActionOption()) {
           StrainsService.addSessionStrain($scope.taxonId, $scope.data.strainSelector)
             .then($scope.getSessionStrains);
         }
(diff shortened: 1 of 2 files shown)


kimrutherford:
> This fix is brittle, since it relies on values that are hard-coded into the template staying in sync
> with values hard-coded into the controller.

I agree that's a bit brittle.  Would this work?:

```html
     <div style="width:280px;">
         <select ng-model="data.strainSelector" ng-change="changed()">
             <optgroup>
-                <option>Add experimental strains for this organism</option>
+                <option value="">Add experimental strains for this organism</option>
             </optgroup>
             <optgroup label="-----------------------------">
                 <option>Type a new strain</option>
```

```perl
@@ -7668,7 +7668,7 @@ var strainPicker = function () {
         newStrain: '',
         strains: null,
         sessionStrains: null,
-        strainSelector: 'Add experimental strains for this organism'
+        strainSelector: ''
       };
 
       CantoService.lookup('strains', [$scope.taxonId]).then(function (data) {
@@ -7708,7 +7708,7 @@ var strainPicker = function () {
       function notActionOption() {
         return (
           $scope.data.strainSelector !== 'Type a new strain' &&
-          $scope.data.strainSelector !== 'Add experimental strains for this organism'
+          $scope.data.strainSelector !== ''
         );
       }
     },
```

jseager7:
I've tested your changes and they work fine, but I think that this change might not be necessary:
```diff
- $scope.data.strainSelector !== 'Add experimental strains for this organism'
+ $scope.data.strainSelector !== ''
```
Because the test for `$scope.data.strainSelector !== ''` is never reached: this case is now handled by logic in `$scope.changed`, because the empty string is falsy:

```js
$scope.changed = function () {
  if ($scope.data.strainSelector && notActionOption()) {
    StrainsService.addSessionStrain($scope.taxonId, $scope.data.strainSelector)
      .then($scope.getSessionStrains);
  }
}
```

Issue #1714, opened 2018-12-11
Use one directive for all genotype lists


jseager7:
(Follow-up from #1691)

The genotype lists on the genotype management page and the **meta**genotype management page have separate templates, controllers and/or directives, despite sharing a lot of functionality:

- Genotype management
  - `genotypeListView`
  - `genotypeListRowCtrl`
- Metagenotype management
  - `genotypeSimpleListView`
  - `genotypeSimpleListRowCtrl`

Ideally, these components should be merged; I think `genotypeListView` or `genotypeListRowCtrl` should be augmented or composed with the radio-button selector functionality, rather than having a separate implementation that handles the radio buttons.

Apparently the rationale for separating the list components was because some dependencies in `genotypeListView` couldn't be fulfilled when the directive was put into the metagenotype page.

PR #1886, opened 2019-05-23
Show genotype background in metagenotype page tables

### LINKED ISSUE #1873: Show genotype backgrounds on metagenotype page
See: https://github.com/PHI-base/curation/issues/22#issuecomment-492281265

jseager7:
Fixes #1873 

The genotype and metagenotype tables on the Metagenotype Management page weren't displaying a column for the genotype background under any circumstances. The background was sometimes needed to distinguish metagenotypes that differed only by their background. This provides a quick fix to that.

The fix definitely isn't ideal, since it causes more divergence between the Genotype Management and Metagenotype Management pages &ndash; the latter already uses a different directive for its genotype tables, called `genotypeSimpleListView` &ndash; my changes are particular to the Metagenotype Management version of the directive, just because it was faster and easier to get implemented. I'm planning to merge the directives on both pages at some point (see #1714), but I haven't figured out a good way to do this yet.

Besides the above problems, the solution seems to work well &ndash; for each table, the background columns are only shown if at least one genotype or metagenotype in that table has a background.

**Note** that the solution uses boolean conversion `!!` on the background field, rather than `null` checking, since editing the background of a genotype to the empty string `""` doesn't set the background back to `null`. I'm not sure if that's meant to be intended behaviour.

- - -

metagenotype-genotype-background

metagenotype-background

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7332,6 +7332,7 @@ var metagenotypeListRow = function (CantoGlobals, Metagenotype, AnnotationTypeCo
   return {
     scope: {
       metagenotype: '=',
+      showBackground: '<'
     },
     restrict: 'A',
     replace: true,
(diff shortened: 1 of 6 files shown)


kimrutherford:
That works well.  Cheers.

PR #1887, opened 2019-05-23
Show organism and strain when editing genotype annotation

### LINKED ISSUE #1767: Display organism and strain when editing genotype annotation
When editing an annotation, or copying the annotation onto a different genotype, there is currently no way to differentiate genotypes that are identical in allele names but differ by strain. We can fix this by appending the strain name to the genotype name in the selector.

jseager7:
Fixes #1767 

Previously there wasn't any way to see the organism or the strain of a genotype when editing the annotation for said genotype. I've fixed that by adding another row to the edit modal:

phenotype-edit-organism

I've restricted the strain name to only display when `CantoGlobals.strains_mode` is true.

@kimrutherford I have a few questions about the changes I've made:

1. Currently the organism name only shows in pathogen-host mode, because I wasn't sure if this would be useful for FlyBase (they didn't want the genotype page filtered by organism, so I'm assuming they can distinguish organisms by allele name). Should the organism name always show in multi-organism mode?

   (If `CantoGlobals.multi_organism_mode` is always true when `CantoGlobals.pathogen_host_mode` is true &ndash; meaning `pathogen_host_mode` is a subset of `multi_organism_mode` &ndash; we only need to check for `CantoGlobals.multi_organism_mode`. Is this the case?)

2. I've chosen to use the `scientific_name` for the organism for display purposes. Should I be using `full_name` instead? In the case of _F. graminearum_, both names are equivalent.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4216,7 +4216,7 @@ var genotypeManageCtrl =
         $scope.app_static_path = CantoGlobals.app_static_path;
         $scope.read_only_curs = CantoGlobals.read_only_curs;
         $scope.curs_root_uri = CantoGlobals.curs_root_uri;
-        $scope.pathogen_host_mode = CantoGlobals.pathogen_host_mode == "1";
+        $scope.pathogen_host_mode = CantoGlobals.pathogen_host_mode;
         $scope.metagenotypeUrl = CantoGlobals.curs_root_uri + '/metagenotype_manage';
 
         $scope.data = {
@@ -4693,7 +4693,7 @@ var genotypeListRowLinksCtrl =
 
         $scope.curs_root_uri = CantoGlobals.curs_root_uri;
         $scope.read_only_curs = CantoGlobals.read_only_curs;
-        $scope.pathogen_host_mode = CantoGlobals.pathogen_host_mode !== "0";
+        $scope.pathogen_host_mode = CantoGlobals.pathogen_host_mode;
 
         $scope.matchingAnnotationTypes = [];
 
@@ -5068,7 +5068,7 @@ var genotypeListViewCtrl =
         }
 
         function getOrganismType(genotypes) {
-          if (CantoGlobals.pathogen_host_mode === "1") {
+          if (CantoGlobals.pathogen_host_mode) {
             var genotype = genotypes[0];
             if ('organism' in genotype) {
               var organism = genotype.organism;
@@ -5666,8 +5666,8 @@ var annotationEditDialogCtrl =
       showEvidence: true,
     };
 
-    $scope.showOrganismName = !! CantoGlobals.pathogen_host_mode;
-    $scope.showStrainName = !! CantoGlobals.strains_mode;
+    $scope.showOrganismName = CantoGlobals.pathogen_host_mode;
+    $scope.showStrainName = CantoGlobals.strains_mode;
 
     copyObject(args.annotation, $scope.annotation);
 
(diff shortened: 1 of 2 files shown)


jseager7:
Just discovered that this isn't working correctly in single organism mode. The organism name still displays. I'll try to fix that.

jseager7:
Turns out this doesn't work in single organism mode because in that mode, `pathogen_host_mode` gets set a string value of zero, which is truthy, rather than a numeric zero or a boolean `false`:

```js
pathogen_host_mode: "0"
```
This is inconsistent with the type of other properties of `CantoGlobals`, which always seem to use boolean values:
```
{
  "alleles_have_expression": true,
  "allow_single_wildtype_allele": false,
  "app_static_path": "http://localhost:5000/static/v1314/",
  "application_root": "http://localhost:5000/",
  "current_user_is_admin": false,
  "curs_root_uri": "http://localhost:5000/curs/ec755ee7e722797f",
  "curs_session_state": "CURATION_IN_PROGRESS",
  "diploid_mode": false,
  "ferret_choose": {},
  "is_admin_session": false,
  "is_admin_user": "",
  "multi_organism_mode": false,
  "pathogen_host_mode": "0",
  "read_only_curs": false,
  "show_genotype_management_genes_list": true,
  "split_genotypes_by_organism": true,
  "strains_mode": false
}
```
The reason why this is happening is because `CantoGlobals` doesn't use loose equality checking for `pathogen_host_mode`, it just copies the value as-is:

```js
this.pathogen_host_mode = $window.pathogen_host_mode;
```
When it should be doing this:
```js
this.pathogen_host_mode = $window.pathogen_host_mode == 1;
```
I'll amend the pull request with that fix.

jseager7:
> I'll amend the pull request with that fix.

Except now I've noticed that, unfortunately, other areas of the code are assuming `pathogen_host_mode` _is_ a string value. 😩 For example:

```js
if (CantoGlobals.pathogen_host_mode === "1") {
```

So they'll break if I change the type of the property. I'll try and fix those at the same time, but this pull request might need some intensive testing if we commit that.

jseager7:
Okay, the bugs above should be fixed now. I've amended the code so it checks the value of `CantoGlobals.pathogen_host_mode` in a more sensible manner:

```js
if (CantoGlobals.pathogen_host_mode) {
```
I've tested in pathogen host mode and single organism mode and it seems everything works fine, providing I haven't missed anything that happens to depend on `CantoGlobals.pathogen_host_mode`. I can't currently test in FlyBase mode, since an error is preventing me from upgrading my local database on that mode to the latest schema version.

kimrutherford:
> Should the organism name always show in multi-organism mode?

I think so.  I probably won't matter much for FlyBase as all their genotypes will be Drosophila (the allele are sometimes from other organisms).

> we only need to check for CantoGlobals.multi_organism_mode. Is this the case?

Yep!

> I've chosen to use the scientific_name for the organism for display purposes. Should I be using
> full_name instead? In the case of F. graminearum, both names are equivalent.

scientific_name and full_name are synonyms.  My plan is to phase out full_name so scientific_name is the right one to use.

> it just copies the value as-is:
> `this.pathogen_host_mode = $window.pathogen_host_mode;`

That seems wrong - `$window.pathogen_host_mode` should be a boolean.  I'll have a go at changing that after merging your pull request.  Note to self, these two lines need changing:
  https://github.com/pombase/canto/blob/137fd553393676c277337c58d170ba83ad1a9839/root/curs/autohandler#L21-L22

Your changes look good.  Thanks.  I'll do a test in FlyBase mode and then merge.

kimrutherford:
> That seems wrong - $window.pathogen_host_mode should be a boolean.
> I'll have a go at changing that after merging your pull request.

After your changes there wasn't much to fix so it's done now.

PR #1898, opened 2019-05-27
Let the organism-selector width be auto

kimrutherford:
Hi James.

Could you give me your opinion of this small change?  I tried it because sometimes things don't look good when genotype management page isn't wide enough.  Can you see any downsides?

wide-selector-1

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1553,10 +1553,6 @@ a:not([href]) {
   color: whitesmoke;
 }
 
-.curs-organism-selector-select {
-  width: 100%;
-}
-
 .strain-selector__container {
   display: inline-block;
 }
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1553,6 +1553,11 @@ a:not([href]) {
   color: whitesmoke;
 }
 
+.curs-organism-selector-select {
+  width: auto;
+  max-width: 300px;
+}
+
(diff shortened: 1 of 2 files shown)


jseager7:
The only potential downside is if `auto` width causes the selector to overflow its container, in the case where we have a long organism name that we haven't remapped to something shorter. The fact that we're using `flex` might already fix this though &ndash; I'll try testing this myself shortly.

jseager7:
> The only potential downside is if `auto` width causes the selector to overflow its container, in the case where we have a long organism name

Having tested this, it's not entirely accurate: the selector doesn't ever overflow, but if it's very wide, it will force the genotype table to wrap below the gene list.

So, It might be better to **set the selector to a fixed width** (something around 300 px) to truncate the longest organism names. This should only be a problem when the organism names contain a lot of strain information (like _Giberella zeae_ PH-1), and hopefully that's not too common. Users can still see the full name when they expand the drop-down anyway.

jseager7:
@kimrutherford I've implemented a change to set `max-width` on the selector in addition to `width: auto`, meaning the selector will always keep its size as small as possible but should never expand beyond the boundaries of its flex container. It might have been better to size the selector according to `flex` sizing rules (to match how its container is sized), but I wasn't sure how that would work.

kimrutherford:
That's great.  Thanks.

Issue #1907, opened 2019-05-30
Canto URLs have spurious forward slashes


jseager7:
I think this problem's been around for a while now, but it might be getting worse. On both the PHI-base servers, Canto has two extra forward slashes after the domain name on its homepage URL. Other links have one extra forward slash.

`https://canto.phi-base.org///curs/13842afb0a578560`

(Note that I just made up the curation session ID here.)

The bug doesn't cause any adverse effects as far as I can tell, it just looks odd. Pombase's Canto is affected too, only not on the homepage for some reason:

`https://curation.pombase.org/demo//curs/fe878952f8154adb`

kimrutherford:
Could you attach your Apache configuration?

jseager7:
Which configuration files specifically? Do you want everything in the `sites-enabled` directory?

kimrutherford:
> Which configuration files specifically?

I'm not sure.  :-)

Could you send an archive of the whole Apache conf directory?  Maybe using send.firefox.com?

jseager7:
> Could you send an archive of the whole Apache conf directory?

Okay, I've done that.

kimrutherford:
Thanks.

The problem is probably this line:
`RequestHeader set X-Request-Base /`

The example in the docs don't have a slash at the end:
https://metacpan.org/pod/Catalyst::TraitFor::Request::ProxyBase

The docs also recommend a full URI so maybe try:
`RequestHeader set X-Request-Base https://canto.phi-base.org`

jseager7:
Okay, that change gets rid of one extra space, thanks. Unfortunately, we still have one more space that appears on the summary page on all versions of Canto, for example:

```
https://demo-canto.phi-base.org//curs/df72946dce1db734
```
Note that this is _only_ on the summary page. All the other pages that I've checked seem to be fine.

kimrutherford:
Hi James.  I can't get to that page: https://demo-canto.phi-base.org/curs/df72946dce1db734

I get "Page not found: curs/df72946dce1db734"

jseager7:
Oh, sorry, I just made that link up. Here's an actual session:

https://demo-canto.phi-base.org//curs/dfd6363574e65f68

kimrutherford:
I can't see any links with a double slash on that page?

jseager7:
Does the URL for that page have a double slash (meaning the one in your browser address bar)? The bug might be browser-specific, which would be weird.

jseager7:
[...]

PR #1921, opened 2019-06-05
Fix assumption that application_root ends in slash

kimrutherford:
Refs pombase/canto#1907


kimrutherford:
Thanks for testing.  The externalTermLinks directive is used by the ontology_term_confirm.html template.  I just checked and it works fine with the changes.

Issue #1891, opened 2019-05-23
Adapt current genetic interactions to allow genotype-to-genotype associations


vmt25:
FlyBase will need to make annotations same gene genotype-to-genotype relationships to describe whether e.g. a full-length wild-type transgene can rescues a loss-of-function 'diploid' mutant.

To explore that, I configured my local test canto so that an 'interaction' data type used genotypes instead of genes. Instead of making the genotype-to-genotype associations I was expecting, Canto tries (and fails) to make a genotype-to-gene associations:

Here are the canto_deploy.yaml lines:

<img width="641" alt="Screen Shot 2019-05-23 at 16 57 02" src="https://user-images.githubusercontent.com/33721108/58267665-ff594280-7d7b-11e9-89c8-31acde569d50.png">

Running canto I can select the type of annotation I configured:

<img width="654" alt="Screen Shot 2019-05-23 at 16 59 50" src="https://user-images.githubusercontent.com/33721108/58267843-43e4de00-7d7c-11e9-8bc1-3e8fdf790e1b.png">

I manage to select the right 'evidence code'/relationship:

<img width="437" alt="Screen Shot 2019-05-23 at 17 02 24" src="https://user-images.githubusercontent.com/33721108/58268019-91f9e180-7d7c-11e9-920e-4f61eed0ea39.png">

But then can only pick genes instead of genotypes, and the annotation fails:

<img width="1059" alt="Screen Shot 2019-05-23 at 17 07 04" src="https://user-images.githubusercontent.com/33721108/58268399-48f65d00-7d7d-11e9-9e12-2a286e5895a6.png">

Would this be easy to fix?

kimrutherford:
> Would this be easy to fix?

This would need changes at the database level because an interaction can only reference genes at the moment.

I think Val has some ideas about how we could implement this so let's have a chat about it on the next Skype call.

vmt25:
Sounds good

ValWood:
Yes let's discuss this on the next call.  it will be really hepful to us to roll the existing GI interface so that  it is don as an extra curation type for multi-allele phenotypes.

Otherwise I often end up having to do the additional annotation to get the specific mutant and phenotype details...
like here:
https://curation.pombase.org/pombe/curs/396866ab84f69995/ro

kimrutherford:
Can we close this in favour of: #1915 ?

kimrutherford:
> Can we close this in favour of: #1915 ?

I've done that now.

Issue #1380, opened 2018-02-08
interaction interface improvements


mah11:
ain't alliteration awesome?

On a tangent from something else, we decided to ask for a big change and a small change to the interface for adding physical or genetic interactions.

Big change: make what is now the "quick add" version the only interface for adding interactions (and therefore not admin-only). It's much more intuitive than stepping through from gene to gene to evidence one page at a time, it's easier to correct directionality, and it'll save space because we won't need the wall of text drawn from the documentation about directionality.

Small change: move the evidence code pulldown so it's between the "gene" and "interacting gene" pulldowns. That way it flows kind of like a sentence in English, e.g. yfg1 affinity captures yfg2.

Let us know if you have problems, questions, pithy observations, etc.

ValWood:
could we do the small change, and punt the larger change?

kimrutherford:
The one thing that's possible in the step by step interface that isn't possible in the dialog box is that you can create multiple interactions at once.  Is that a big loss?  If not, swapping to using the dialog box is just a matter of changing what the link does.

![interactors](https://user-images.githubusercontent.com/90474/38784531-5db2783a-4167-11e8-8756-7806fa290a00.png)

ValWood:
Personally I don't mind not being able to create multiple interactions at once, since I always found that a bit confusing (and there are not so many to curate).
Anyone else?

mah11:
agree, no loss - it's easy enough to use the duplicate/edit feature

Antonialock:
no objections

kimrutherford:
OK, I've removed the quick links for interactions and the main links now pop up the quick add dialog.

I've released the change in the main Canto.  Could you check that it's doing what you wanted?

mah11:
Yes, working nicely. Thanks!

kimrutherford:
> Yes, working nicely. Thanks!

Great.  Thanks for checking.

PR #1959, opened 2019-06-17
Change metagenotype table to handle interactions

kimrutherford:
Hi James.  Another week, another schema change.  :-)

To implement #1891,I'm suggesting the we make the metagenotype table more general by renaming the pathogen_genotype_id and host_genotype_id columns.  And by adding a "type" column the type would be constrained to "pathogen-host" or "interaction" (see https://github.com/pombase/canto/blob/tickets/ticket-1891/etc/curs.sql#L77).

For #1891 we need to store a genotype to genotype interaction in the database and then connect that to phenotype term and an evidence code.  I'd prefer to re-used the metagenotype table to hold these interactions and to reuse as much metagenotype code as possible.

I'm not planning to merge this pull request yet but I wanted to make a PR for discussion.  I'm going to try implementing the basics of the interaction interface before I think about merging to make sure that these changes make sense.


jseager7:
While I agree with the idea of generalising a metagenotype to cover organisms in any role (not just pathogen and host), I'm slightly concerned about how generalising this might harm the readability of the front-end code, in the sense that it could make it more difficult to reason about which organism is the pathogen and which is the host when working on the view model, and so on.

This is important in the case of PHI-Canto, since generally the pathogen data should be displayed 'first', meaning further to the left than the host data.

I don't know what the naming conventions are going to be, but I guess we could come up with some sort of de-facto mapping that `genotype_a` is the pathogen and `genotype_b` is the host. It might not be a problem in practice, and I'm not sure of a good way to solve this issue yet anyway. Maybe it will be more apparent once the code is changed.

jseager7:
As for the changes required, there is at least some code that is using dynamic naming based on things like the organism type, for example:

```js
$scope.getOrganismName = function (type) {
        var metagenotype = $scope.metagenotype[type + '_genotype'];
        var name = metagenotype.organism.full_name || '';
        return name;
      };
```
Anything like this will need changing, but I think the general solution is better since hopefully we can avoid having to account so much for domain-specific naming in the code.

kimrutherford:
> As for the changes required, there is at least some code that is using dynamic naming based on things
> like the organism type, for example:

Thanks for that example.  I hadn't thought about that case.  I'll do a thorough check of the JS code before we merge the changes.

kimrutherford:
> I'm slightly concerned about how generalising this might harm the readability of the front-end code

I'm hoping we won't need to change any of the current front end code.  The Perl webservices can continue to accept the same requests and return the same responses in the PHI case.  Because we'll have a "type" field in the metagenotype table the server code will always know if the metagenotype it's looking at is a pathogen-host or genotype-genotype interaction.

So now I've thought about it hopefully this won't be a problem:
> As for the changes required, there is at least some code that is using dynamic naming based on
> things like the organism type, for example:

jseager7:
> Because we'll have a "type" field in the metagenotype table the server code will always know if the metagenotype it's looking at is a pathogen-host or genotype-genotype interaction.

Now that I've actually bothered to read the issue, I realise that the view logic for these two cases of interaction has almost nothing in common, and really the only thing we're abstracting is the data layer. In that case, I think you're right about the case of not needing many front-end changes.

My only concern was whether these two cases will have some logic in common at the controller or service level. The `Metagenotype` service doesn't seem to make any assumptions as far as I can tell, so that should be fine. Maybe the only issue is the components used on the PHI-Canto Metagenotype Management page (like `metagenotypeListRow` and so on) &ndash; right now they only cover the pathogen-host case. I think it would be better to either:

* rename these components to make their pathogen-host functionality explicit, since they are currently not general enough to handle all types of metagenotype; or

* generalise the controllers to handle any type of genotype, and make the domain-specific labelling configurable (or, alternatively, assign their templates dynamically).

This is presuming you'd ever want a way in FlyBase to list the metagenotypes in a table, in the same way as is done with PHI-Canto. If that's not likely, then we can probably get away with not even renaming the components.

kimrutherford:
>The Perl webservices can continue to accept the same requests and return the same responses
> in the PHI case. 

I forgot that I've done some of that already.  Because of this change, creating new metagenotypes works after the schema update:

https://github.com/pombase/canto/commit/b549481c17cae4d2bd56f4f248594002a8159dad#diff-a214cd50f78d3efb2cddc1729a1cba53R502

Creating a PHI metagenotype annotation still works.  And the aliases here mean that the metagenotype annotations display correctly:
https://github.com/pombase/canto/commit/b549481c17cae4d2bd56f4f248594002a8159dad#diff-acd857e045721043ff88245ce36d6b88R162

Some of the server code still assumes that all the entries in the metagenotype tables are pathogen-host interactions.  I'll fix that before we merge.

jseager7:
I'm going to push any simple fixes to this pull request instead of opening issues. I've just fixed an error in commit 9c82ec7, in the case where there's only one organism on the pathogen or host genotype management pages.

jseager7:
Another thing I've noticed is that the 'Quick links' section is being shown permanently now, even when I'm not in advanced mode. All it shows is a link to physical interaction, though. Not sure if this is just something that's been enabled as a convenience for development?

image

kimrutherford:
> Another thing I've noticed is that the 'Quick links' section is being shown permanently now,
> even when I'm not in advanced mode. All it shows is a link to physical interaction, though. Not
> sure if this is just something that's been enabled as a convenience for development?

It's sort-of for development. But I also wanted to see if the curators were interested in having a quick link to make an interaction rather than navigating to a genotype details page first.

>  I haven't been able to test the genetic interaction workflow myself, because I'm not sure how to enable it.

There isn't a long workflow for interactions any more, we can only create interactions via the dialog: https://github.com/pombase/canto/issues/1380

jseager7:
> I haven't been able to test the genetic interaction workflow myself, because I'm not sure how to enable it.

I've just remembered that I didn't have genetic interaction enabled in the `canto_deploy` file, so that will be why I couldn't see it. We disabled it for PHI-base a long time ago because nobody in the team thought we'd need it.

kimrutherford:
Thanks very much for testing and for the fixes.  I'll do more testing too before merging.  It will be a while before I do that as I need to fix things so that we can store an interaction without a phenotype.  Currently a phenotype is required.

Also I need to write the code to convert old gene-gene interactions in to genotype-genotype interactions when the database is upgraded.  And for PomBase pombase/pombase-chado#739 and pombase/website#1348 need to be done.  It will take a while!

I've been rebasing this branch occasionally so that the final merge won't be too painful.

ValWood:
I haven't seen any genetic interactions yet.

As we were working on the new geentic interactions interface we were thinking how much that traditional genetic interactions have in common with multi-organism phenotypes.

It is likely that  both interfaces could benefit from input from the workflow of the other, but I haven't been able to  pull this apart in my head yet , and we have a good working solution for both.

Because researchers in the pathogen community are  usually concentrating on the inter-organism interactions they don't tend to look at "within species genetic interactions". It could be enabled later if required.

kimrutherford:
Hi James.

I'm closing to close this pull request because it's getting out of hand and the branch name doesn't make sense any more.  I've made a new branch (https://github.com/pombase/canto/tree/genotype-interactions) that contains the schema change and all the other changes that depend on it.

PR #1964, opened 2019-06-27
Small improvements to navigation buttons

jseager7:
This PR makes some small changes to navigation buttons, mainly in pathogen-host mode, to make them more sensible and/or consistent with the rest of the application:

* Remove the right arrow from the 'Make metagenotype' button on the Metagenotype Management page: the button doesn't actually navigate forwards, so it shouldn't look like a navigation button.

* Rename the back button on the metagenotype details page from 'Back' to 'Back to metagenotypes', which is more consistent with the Genotype Management page.

* Put the 'back' button on the Metagenotype Details page above the metagenotype list, which is more consistent with the Genotype Management page.

### DIFF
--- a/root/curs/metagenotype_page.mhtml
+++ b/root/curs/metagenotype_page.mhtml
@@ -62,13 +62,11 @@ Description
 
 <div class="clearall"/>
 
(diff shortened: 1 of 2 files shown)


kimrutherford:
Looks good to me.  Thanks!

Issue #1955, opened 2019-06-14
Tidy up the genotype table


kimrutherford:
I've been experimenting with the genotype table.  I think there are too many vertical and horizontal lines at the moment.

Is this an improvement?:

![new-style-genotype-table-4](https://user-images.githubusercontent.com/90474/59507572-7a3af680-8eff-11e9-9cc4-721ce22231f3.png)

We've talked about hiding the gene links.  Perhaps we could hide them until you mouse over the genotype?:

![new-style-genotype-table-5](https://user-images.githubusercontent.com/90474/59507850-5e842000-8f00-11e9-8de8-9b670848dd0a.png)

ValWood:
Oh yes. much.

mah11:
fine with me as long as long things wrap sensibly, so it's easy to tell an continuation apart from the next allele in the genotype

e.g. something like this would be ok:

    cdc17-1-175/ΔNLS-GFP(N-terminal 175 amino acids of cdc17, with 
                 KKK62AAA mutations, fused to GFP)[Not assayed] 
    cds1Δ

but this would annoy:

    cdc17-1-175/ΔNLS-GFP(N-terminal 175 amino acids of cdc17, with
    KKK62AAA mutations, fused to GFP)[Not assayed] 
    cds1Δ

kimrutherford:
Instead of line wrapping, there's also the option of hiding the end of the text until you mouse-over or click:

![new-style-genotype-table-6](https://user-images.githubusercontent.com/90474/59578755-e3985080-911d-11e9-87fc-1c4b1bb73558.png)

after clicking:

![new-style-genotype-table-7](https://user-images.githubusercontent.com/90474/59578789-0cb8e100-911e-11e9-8560-c29f9b174581.png)

kimrutherford:
> I've been experimenting with the genotype table. 

The test Canto has the new look.

ValWood:
I will give it a whirl later. It might also be useful to make a change we discussed a long while ago

[Wild Type Product Level] ->[WT product level]
(also discussed on PHI-BAse call)

jseager7:
The solution here is the same one that was discussed in #1520, and that issue also got stuck on the problem of what happens when genotype names are long enough to wrap onto the next line. If we can solve that here, we can close #1520 too.

@kimrutherford with regards to text appearing on mouse-over and click: does showing the gene names on mouse-over cause the size of the table to change, or does the table always account for the width of the hidden gene names?

[...]

PR #1978, opened 2019-07-12
Tidy up the genotype table

kimrutherford:
I've made a branch and pull request for this so I don't misplace the changes I've made.  It's not ready for review or for merging.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -414,6 +414,10 @@ table.curs-genotype-list tbody {
   padding: 3px;
 }
 
+.curs-genotype-locus-alleles {
+  white-space: nowrap;
+}
+
 .annotation-stats {
   clear: both;
   padding-top: 30px;
@@ -734,10 +738,18 @@ div.clearall {
   margin-left: 1em;
 }
 
-table th.curs-genotype-checkbox-column,
-table td.curs-genotype-checkbox-column {
+table tr th.curs-genotype-checkbox-column,
+table tr td.curs-genotype-checkbox-column {
   border: 0px solid #bbb;
   background-color: white;
+}
+
+tr.selected .curs-genotype-allele-gene {
+  visibility: visible;
+}
+
+tr .curs-genotype-allele-gene {
+  visibility: hidden;
 }
 
 .curs-genotype-list-instructions span {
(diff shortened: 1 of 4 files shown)


kimrutherford:
I forgot to say these changes are for issue #1955.

kimrutherford:
Hi James.  I've updated this pull request to master.  Could you try it out when you get a chance?  This PR changes (hopefully improves) the like of the genotype table by removing horizontal lines.  As described in #1955.  It also makes the template slightly simpler which is a bonus.

jseager7:
I've tested this locally and there don't seem to be any major problems.

There's the minor issue that the table layout is no longer consistent with what shows on the metagenotype management page (because they don't share directives), which could be slightly confusing for users but nothing serious.

One thing that's different to `master` is that the 'new notes' link (that appears when you've added a note to a genotype) is hidden along with the gene name until you hover over the row. This can lead to a lot of seemingly redundant empty space in some rows:

Not hovered:
image

Hovered:
image

On `master`, the 'new notes' link is always visible, hovered or not:

image

Was this supposed to happen?

jseager7:
One other thing I was wondering, in the interest of saving space, is whether it would be better to have the gene name shown in a popup or tooltip rather than rendered into the row. That way we wouldn't need to have the space constantly reserved just for the event that the user doesn't remember or recognise the gene involved in the genotype.

We could have this popup / tooltip appear when the genotype name is hovered, when the row is hovered, when either are clicked, or we could implement all of those options at once.

This solution would be less suitable if curators often name their genotypes in such a way that it's not clear what the underlying gene is, but I don't know how common that is.

kimrutherford:
> One thing that's different to master is that the 'new notes' link (that appears when you've added a
> note to a genotype) is hidden along with the gene name until you hover over the row.

Thanks for spotting that.  I think it's best it the "new notes" link is always visible.  I'll change it.

> is whether it would be better to have the gene name shown in a popup or tooltip rather than rendered into the row.

Possibly, but the gene name is a link.  So the pop-up would need to have a link in it.

jseager7:
> Possibly, but the gene name is a link. So the pop-up would need to have a link in it.

I guess if we're planning to make the 'new note' button visible again then the excess of blank space might not be as much of an issue. It's probably only worth worrying about this if and when the screen width becomes constrictive, because adding more pop-ups (in addition to the action one we already have) could make the interface look crowded.

kimrutherford:
> I think it's best it the "new notes" link is always visible. I'll change it.

I've fixed that. We end up with quite a bit of white space but I think the table is tidier.

kimrutherford:
I'm going to merge this.  We can fix any problems later.

PR #1998, opened 2019-08-06
Redesign summary page gene list to not use tables

### LINKED ISSUE #1981: Quick links overlap gene list in advanced mode
image
image

@jseager7 please could you look into this as discussed. Thanks

jseager7:
(Fixes #1981)

This pull request redesigns the gene list on the curation summary page to be a simpler, two column layout. Pathogens are placed in the left column and hosts are in the right column. The list will collapse to a single column view when there isn't enough screen width.

Here's how it looks:

image

@kimrutherford I haven't tried to run the genes across two columns yet &ndash; I'm planning to try that next.

One thing I was wondering about is whether we want to show organisms that have no genes. The old table did this, but I'm not sure if there's much point, since the list is about annotating genes, and you can't annotate a gene if you don't have any for the organism.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1627,3 +1627,18 @@ a:not([href]) {
 .curs-pub-finished-message-to-curator  {
   white-space: pre;
 }
+.curs-summary-gene-box {
+  overflow: auto;
+}
+.curs-summary-gene-column {
+  float: left;
+  margin-right: 10px;
+  padding-right: 10px;
+  margin-top: 10px;
+}
+.curs-summary-gene-organism-box {
+  margin-bottom: 8px;
+}
+.curs-summary-no-gene-text {
+  color: #737373;
+}
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1627,7 +1627,7 @@ a:not([href]) {
 .curs-pub-finished-message-to-curator  {
   white-space: pre;
 }
-.curs-summary-gene-box {
+.curs-summary-gene-container {
   overflow: auto;
 }
 .curs-summary-gene-column {
@@ -1636,9 +1636,12 @@ a:not([href]) {
   padding-right: 10px;
   margin-top: 10px;
 }
-.curs-summary-gene-organism-box {
+.curs-summary-organism-container {
   margin-bottom: 8px;
 }
 .curs-summary-no-gene-text {
   color: #737373;
+}
+.curs-summary-gene-name {
+  display: block;
 }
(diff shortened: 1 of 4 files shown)


kimrutherford:
> One thing I was wondering about is whether we want to show organisms that have no genes.

I think it's good to keep the organisms with no genes as this page is the only place where you can see all the organisms in the session.

kimrutherford:
> Here's how it looks:

Looks good!  I haven't tested locally yet.  Let me know when you're done with changes and then I'll give it a go.

jseager7:
@kimrutherford Thanks for the feedback. One other thing I was wondering was that the old code had this function, which stripped bracketed text from the organism names (to keep them from getting too long):
```js
$scope.tidyName = function (name) {
  var pos = name.indexOf("(");
  if (pos > -1) {
    name = name.substring(0, pos);
  }
  return name;
}
```
But even with this function removed, I still don't see brackets appearing again in the organism names &ndash; even for ones that aren't in our PHI-base organism list (meaning they're not being automatically renamed either). Maybe the Perl service is passing back an abridged name already? Just wondering if you had any ideas, so I can be sure that this function is redundant.

kimrutherford:
I've had a look in the server code and I can't see any place where the brackets are removed.  I think it's safe to delete tidyName().

jseager7:
@kimrutherford I tried out spreading the gene list across more than one column for each organism, but unfortunately the column layout rule applies to all organism groups, rather than affecting each one individually. That means if one long gene name forces the layout to one column (rather than two or more), every other group does the same thing.

There might be other ways to achieve this effect, but I'm not sure it's worth spending the time on it until we have actual examples of sessions where there are so many genes that the annotation tables get pushed far off-screen.

I expect it will be easier to tweak the overall page layout to save screen width and height, for example: using more flex layout instead of bootstrap columns; reducing font sizes to 14px (to be consistent with other links); moving the workflow links up into the same row as the 'Annotate genotypes' section; and so on.

I'm marking the pull request as ready to be merged, but please let me know if you want anything else changed.

kimrutherford:
Thanks James.  I've tried it locally and I have two suggestions: maybe make the font for the gene names a little bit smaller and maybe indent the gene list a little.

```
.curs-summary-gene-name-container {
    display: block;
    font-size: 95%;
    padding-left: 0.5em;
}
```

I'm going to merge the pull request.  My suggestions won't make much difference.

jseager7:
@kimrutherford I noticed that the font size wasn't consistent with some other links. The 'Quick links' section also has the same problem (that's at 16 px rather than 14.4 px). Did you intend for Canto's 'base' font size to be around 14 px? Because I could look at doing a more general fix if that's the case.

PR #2005, opened 2019-09-05
Add inline help text for genotype annotation links

### LINKED ISSUE #1658: Add help text for summary page genotype heading
Pull request #1649 added a new heading on the summary page to separate the genotype links from the gene links. However this new heading doesn't have any help text associated with it, whereas the gene heading does. Is it worth adding any help text?

jseager7:
(Fixes #1658)

This pull request adds some missing inline help text for the 'Annotate genotypes' heading on the summary page. This keeps the heading consistent with the 'Annotate genes' heading.

I'm marking this pull request as a draft, since the help text might need some further edits, but we have the option to merge if we need to.

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -2029,9 +2029,10 @@ help_text:
       set of annotation options
   choose_genotype:
     inline: >
-      Phenotypes annotations are connected to genotypes in Canto. Required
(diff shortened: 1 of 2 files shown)


jseager7:
The text has been edited and agreed now, so this PR is ready to merge.

PR #2006, opened 2019-09-05
Use a custom modal for delete confirmations

### LINKED ISSUE #1589: Fix inconsistent dialog buttons and style
The delete annotation / delete genotype confirmation dialogs should use the same style and button order as the edit dialogs.

confirm-delete-1

edit-dialog-1

jseager7:
(Fixes #1589)

This pull request adds a custom modal that can be used for confirming the 'Delete' action on annotation table rows. The modal function is `openDeleteDialog`, with corresponding controller `DeleteDialogCtrl` and template `delete_dialog.html`.

Here's how it looks:

image

The modal is essentially a specialisation of `simple_dialog.html` with different buttons (and extra handlers for the extra buttons). I think this is a bit wasteful, since it's probably possible to generalise the existing simple dialog to allow custom buttons &ndash; although passing the handlers around could get complicated.

@kimrutherford I can try to generalise the simple modal if you want me to, but feel free to merge this if you don't feel the extra work is worth it.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -6629,6 +6629,18 @@ var annotationTableRow =
             newAnnotation, $scope.featureFilterDisplayName,
             true, true);
         };
+        
+        $scope.confirmDelete = function () {
+          var modal = openDeleteDialog(
+            $uibModal,
+            'Delete Annotation',
+            'Delete Annotation',
+            'Are you sure you want to delete this annotation?'
+          )
+          modal.result.then(function () {
+            $scope.deleteAnnotation();
+          });
+        }
 
         $scope.deleteAnnotation = function () {
           loadingStart();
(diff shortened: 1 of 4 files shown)


jseager7:
It looks like Travis-CI failed because of a bad gateway from the NCBI E-Utilities:

```
t/60_load_util.t ........................... 1/16 Couldn't read from http://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=pubmed&retmode=xml&retmax=99999&id=2120045: 502 Bad Gateway
# Looks like your test exited with 255 just after 4.
```

Not sure if this is a temporary issue, or if the test needs fixing. It's affecting more than one PR at the moment.

kimrutherford:
> I can try to generalise the simple modal if you want me to, but feel free to merge this if
> you don't feel the extra work is worth it.

I think a specialised delete dialog makes sense.

> Not sure if this is a temporary issue, or if the test needs fixing.

It's just temporary.  It has been happening more often recently so we may need to change the code to do a mock request instead.

PR #2007, opened 2019-09-05
Add genotype management shortcuts to feature pages

### LINKED ISSUE #1896: Add a link to the genotypes page from here:
access to genotypes page

I often want to add a genotype from the gene page. 

In fact if I am curating a complicated paper with lots of genotypes I often need to go to the gene page view to see only the ones I have already added for a specific gene.

To do this I need to go back tot he summary and then to the genotype page. I think it would be useful to have a link to 
"genotype management"  under the genotypes list (empty here), so you can go directly to add more.

It doesn't seem a lot, but it  could add many  extra clicks  per session...

jseager7:
(Fixes #1896)

This pull request will add a forward navigation button (to complement the back navigation button) to the gene and genotype detail pages.

Currently, the PR creates a new directive `genotypeAndSummaryNav` that keeps these buttons together. The genotype management button is able to link to the correct genotype management page based on a 'role' binding (which acts on the values 'pathogen' or 'host' and otherwise defaults to the standard genotype management page.

This PR is currently a work in progress, due to an unresolved problem preventing this from working on the genotype page: see https://github.com/pombase/canto/issues/1896#issuecomment-528373625.

### DIFF
--- a/root/curs/gene_page.mhtml
+++ b/root/curs/gene_page.mhtml
@@ -130,7 +130,7 @@ Genotypes for <% $gene->display_name() %> in this session
 
 <div class="clearall"/>
 
-<a href="<% $summary_url %>" type="button" class="btn btn-primary curs-back-button">&larr; <% $finish_text %></a>
+<genotype-and-summary-nav role="<% $gene->organism_details()->{pathogen_or_host} %>"></genotype-and-summary-nav>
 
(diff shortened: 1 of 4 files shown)


jseager7:
The problems with the genotype page are resolved now (due to a536f94) so I'm marking this as ready for testing / merging.

PR #2030, opened 2019-09-24
Replace extension hint message with input placeholder

### LINKED ISSUE #1778: Unhelpful extension message for small ontologies (i.e. dropdown menu)
In the extension edit dialog, we should have a better message in the case where the terms are in a dropdown rather than an autocomplete.

"(Start typing a description)" doesn't make sense here:

relation-dropdown-message-1

jseager7:
Fixes #1778

This pull request removes the message in the extension dialog that instructs the user to 'Start typing a description' when the input isn't text editable (not an `<input>`). Instead, that message is now provided as the placeholder text for the `<input>`; previously the placeholder was only a term name suggestion.

Here's how the placeholder looks now:

image

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -6809,6 +6809,7 @@ var termNameComplete =
         $scope.typeMatch = re.exec($scope.annotationTypeName);
 
         if ($scope.typeMatch) {
+          $scope.placeholder = 'Start typing a term name, e.g. ';
           var split = $scope.typeMatch[1].split(/\s*\|\s*/);
           var promises =
             $.map(split,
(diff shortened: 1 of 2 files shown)


jseager7:
@kimrutherford I was just wondering what happens in the case where the ontology lookup (`CantoService.lookup`) fails to find any term suggestions. Presumably then the placeholder would just read "Start typing a term name, e.g. ". I could prevent that by only appending the "e.g." once the promises resolve, but is this a risk worth worrying about?

kimrutherford:
> I was just wondering what happens in the case where the ontology lookup
> (CantoService.lookup) fails to find any term suggestions

It is looking up using the IDs from the configuration.  If there is a server error the "then()" won't happen and the placeholder won't be set at all.  We should probably at least "console.log()" that case.

If any of the IDs return no results, it's a configuration problem.  I'm not sure what's best in that case.  We could say "can't find 'SO:000123' " in the placeholder or just put the ID in the placeholder.  Hopefully this is an edge case that doesn't happen in production.

kimrutherford:
Thanks James.  This looks good too me.

PR #2048, opened 2019-10-01
Improve clarity of front page quick links

### LINKED ISSUE #2042: quick links line wrapping
This isn't a big deal right now, because its an admin only option (I think? or is it more generally avaiable now?).

Anyway the curation  type  "pathogen host interaction phenotype " line wraps so it looks like  2 annotation types:

jseager7:
Fixes #2042 

This pull request affects the list of quick links that are shown on the front page when in advanced mode. Previously, the number of links could be ambiguous when long link names wrapped onto multiple lines:

image

I've addressed this by adding bullets to the list, and capitalising the first letter of every list item. Now it's more obvious how many list items there are, even when there's an excessive amount of line wrapping:

image

This change affects every version of Canto, but it's only really a problem for PHI-Canto due to the long annotation type names.

### DIFF
--- a/root/curs/front_page_quick_links.mhtml
+++ b/root/curs/front_page_quick_links.mhtml
@@ -7,13 +7,15 @@
     Quick links
   </div>
   <div class="curs-box-body">
+    <ul class="curs-quick-links-list">
 % for my $annotation_type (@annotation_type_list) {
-    <div>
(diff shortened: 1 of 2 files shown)


jseager7:
I'm going to merge this now since it's a fairly minor change, even if it affects all versions of Canto. I've tested in all three modes and didn't see any problems.

PR #2050, opened 2019-10-02
Add option for unknown expression level

### LINKED ISSUE #2046: sometimes  WT expression level is unknown
Sometimes  the  WT pathogen gene is expressed in the host (either by some transgenic construct  or by adding isolated protein to host cells.

In these cases, I don't think it is always possible to sensibly say anything about expression level. So I want to be able to do  WT gene/ expression level unknown.

I'm sure we discussed allowing this for PHI-Canto at some point?

jseager7:
Fixes #2046 

This pull requests adds an option to specify the allele expression level as 'Unknown'. I think the option is only currently needed for pathogen-host mode, so I've restricted it to that mode in the template. I've tested allele creation with the new expression type and everything seems to work fine.

image

image

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -3330,6 +3330,7 @@ var alleleEditDialogCtrl =
     };
 
     $scope.userIsAdmin = CantoGlobals.current_user_is_admin;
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks James.  That's great.  Ideally the expression levels should be in the config file but that's a job for another day.

PR #2062, opened 2019-10-07
Add missing ng-click to delete host button

### LINKED ISSUE #2041: Small bug in gene "deselector"
If I have used a gene or organims in an annotation it should not be possible to 'deselect'.

this is probably the case, but here, I have used rice in a metagenotype annotation,  However, the red x is present so I could theoretically delete it (I did not try this).

bug X

(an unimportant aside but these x's aren't very aesthetically pleasing)

jseager7:
Fixes #2041

This pull request fixes a regression in the host organism deletion button that removed the click handlers for deleting the organism.

@kimrutherford I would have pushed this straight to `master`, but I wasn't sure if modifying templates requires a version bump in `canto.yaml` for cache-busting. I'm happy to amend this pull request to add that change.

### DIFF
--- a/root/static/ng_templates/edit_organisms_table.html
+++ b/root/static/ng_templates/edit_organisms_table.html
@@ -40,11 +40,12 @@
           <td ng-if="o.genes.length == 0">
             <button ng-if="canRemoveHost(o)"
               title="Remove host"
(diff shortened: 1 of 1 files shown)


kimrutherford:
Thanks for catching that James.

Issue #2072, opened 2019-10-14
organism/strain column


ValWood:
I notice metagenotype table now has

Magnaporthe oryzae (strain: Guy11)

We should

- [x] 1 change the column heading to 

species (strain)
and
chance the table text to
Magnaporthe oryzae (Guy11)

All users will be used to seeing species name (strain in parentheses)  in this context so we don't need to be explicit

- [x] 2. The  pathogen and host-only tables should follow the same convention

ValWood:
- [ ] "Unknown strain used"
throughout change text to just "unknown strain"

jseager7:
Related: #2037

jseager7:
> "Unknown strain used" throughout change text to just "unknown strain"

@kimrutherford would we need changes in the database to amend the strain names for existing genotypes that have been created using the 'Unknown strain used' option? It's one of the default options in the strain picker:

![image](https://user-images.githubusercontent.com/37659591/67204628-d65ddc00-f405-11e9-869b-3c578c9e3489.png)

jseager7:
@ValWood The changes to the metagenotype table are ready now. As part of the changes I've also implemented abbreviations for the genus, and I've removed the bold styling on the species names, since I wasn't sure if the visual emphasis was necessary. Which style do you prefer? (see images below)

![image](https://user-images.githubusercontent.com/37659591/67205065-d1e5f300-f406-11e9-99f7-5483d7356120.png)

![image](https://user-images.githubusercontent.com/37659591/67205108-e75b1d00-f406-11e9-964e-c4534303b276.png)

ValWood:
I think the table will look less busy generally if it isn't bold maybe? We can always change it later, but let's try "not bold".

Looks good, this will be much better combined.

PR #2082, opened 2019-10-22
Improvements to organism display in metagenotype table

jseager7:
References #2072 

This PR makes a few small changes to the metagenotype table on the Metagenotype Management page, in order to reduce the space used, and to make the styling more consistent with other annotation.

Note that this PR uses a new AngularJS filter called `abbreviateGenus` which is used to abbreviate the genus of a scientific name to a single character, for example: _Arapidopsis thaliana_ &rarr; _A. thaliana_. The annotation tables don't use this formatting yet, but there's another PR in the works that will implement that.

Before:

image

After:

image

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1659,5 +1659,5 @@ a:not([href]) {
 }
 
 .curs-strain-name-inline {
-  display: inline-block;
+  display: block;
 }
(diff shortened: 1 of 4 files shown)


kimrutherford:
Hi James.  Looks good.

I suggest this tweak in style.css:
```css
 .curs-strain-name-inline {
-  display: inline-block;
+  display: block;
 }
```

to make sure that the stain is on it's own line, preventing this:

metagenotype-style-fix-1

jseager7:
> to make sure that the stain is on it's own line, preventing this:

I don't want to be pedantic, but that isolate of [_T. mosaic_ virus](https://www.uniprot.org/taxonomy/1289472) probably shouldn't ever be used in PHI-Canto because it's subordinate to the species rank. Shouldn't it have been remapped to the [species-level entry](https://www.uniprot.org/taxonomy/431317) of _T. mosaic_, which doesn't have all that parenthetical information?

Otherwise I agree that block display is probably going to be more consistent. It's not really wasteful on space either, since we're already wasting space by not having pop-up links on the metagenotype table.

Issue #2037, opened 2019-09-30
Really hard to navigate due to  table width


ValWood:
I think this is going to be a real problem for lots of sessions.
![too wide](https://user-images.githubusercontent.com/7359272/65895376-46e38100-e3a3-11e9-989c-bb5accafca57.jpg)

ValWood:
I can't remember which suggestions were discussed int he past but here are some for discussion (probably again)

1 Combine organism and strain column.
2. Shorten organism name to 
M. oryzare and only keep more letters in genus if further disambiguation is required

ValWood:
3. for alleles of type "other" which tend to be the ones with long text descriptions, only show the first few characters of the text.
Here this would be MoChai1-NSP(SPΔ(res...)

jseager7:
It's almost inevitable that we're going to run into width issues using tables, because we can only add more properties to each concept by extending the width. About the best we can do, like you've said, is to collapse columns together (another option here is moving the PHIPO term ID into the 'Term name' column).

Unfortunately collapsing columns removes the main advantage of tables: the ability to easily compare values in categories of data. We probably need to consider which columns this comparison is and isn't useful for, and then collapse the columns where this comparison isn't so useful.

I feel like I've asked this before, but I'd be interested to know what the main reason is for keeping tables. Is it to allow easy comparison within categories of data? For example, is it useful to be able to scan down the Gene or Genotype columns? One thing that might be useful to think about is which columns you find yourself using as an 'index', i.e. to help locate data when you're looking at the table.

ValWood:
Good points. Personally I think we need to see the table as we make annotations. I often scan to find the closest annotation to copy edit, to see if I used an allele before, to sanity check the whole contents. I think if I need to filter it would drive me crazy. It would be useful to assess though...it might be habit
@mah11  @Antonialock  @CuzickA  ?

I still that if we can make small changes to the table without loss of info  we should (until we think of a better solution).  Most tables are fine  for PomBAse Canto on a laptop screen.  However, the addition of a strain column, and longer gene IDs have pushed  it over the edge for PHI-Canto as soon as you have  something longer than normal like "allele type other"

Options 1&3 would help. I don't envisage we would ever add another column now (unless for a completely new application).

jseager7:
> Most tables are fine for PomBAse Canto on a laptop screen. However, the addition of a strain column, and longer gene IDs have pushed it over the edge

[...]

PR #2086, opened 2019-10-23
Improvements to display of annotation tables (multi-organism)

jseager7:
References #2037

This PR makes several changes to how the annotation tables are displayed in multi-organism / pathogen-host mode:

* merge the 'Strain' column with the 'Organism' column, and rename the column to 'Species (strain)';
* remove bold styling on organism and strain names;
* apply italic formatting to scientific names;
* ensure that strain names do not wrap, to prevent silly line breaks after a hyphen (in PH-1, for example).
  * Note I did this by applying the `inline-block` display mode to the `<span>` that holds the strain; if this disables any wrapping on long multi-word strains, that might not be ideal. Perhaps a better fix is to use pre-processing in the controller (or a filter) to substitute punctuation for their non-breaking variants, where possible.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1080,11 +1080,6 @@ fieldset.curs-genes span {
 .curs-annotation-table {
   
 }
-
-.curs-annotation-row-strain {
-  white-space: nowrap
-}
-
 .curs-annotation-table .term-warning {
   color: #f00;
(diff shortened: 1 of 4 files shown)


kimrutherford:
Thanks James.  It looks good to me and I couldn't see any problems in single organism mode.

Issue #2101, opened 2019-11-06
Cannot differentiate same gene from different species  when selecting genes for extensions


ValWood:
See:

<img width="703" alt="Screenshot 2019-11-06 at 12 51 27" src="https://user-images.githubusercontent.com/7359272/68312667-4e393100-00ab-11ea-9801-ac984e479161.png">

ValWood:
Note that, even after you selected you can't check you have the correct one either...

kimrutherford:
Do you have the two UniProt IDs handy so I can reproduce the problem locally?

ValWood:
I1S9X9
and
Q51WZ9

in 9133cb6f3978f78d/edit_genes

kimrutherford:
Do we need to add the organism after the gene name?

![annotation-edit-gene-1](https://user-images.githubusercontent.com/90474/68555421-9339da80-0492-11ea-90df-597a77c144ea.png)

jseager7:
@ValWood are you specifically talking about the gene drop-down on the annotation extension dialog?

![image](https://user-images.githubusercontent.com/37659591/68580086-8787ed00-046d-11ea-9e2b-a0197fec0a35.png)

Because the image in your initial comment shows the gene drop down on the GO annotation dialog:

![image](https://user-images.githubusercontent.com/37659591/68580156-ad14f680-046d-11ea-85cb-8c74e724ff43.png)

It looks like the GO annotation drop-down is showing genes for all organisms, rather than the organism being annotated (this seems a bit counter-intuitive to me, in the example above I'd expect to only see the genes TRI5 and URA7, since these are the only genes in the session for _F. graminearum_). Not sure if we have an issue open for this already.

As for the annotation extensions, most of the extensions that involve a gene ID in PHI-Canto are related to an assay (assayed_using) &ndash; in this case, does it ever make sense to assay using a gene from a different organism? Is that why you needed to select a gene from another organism?

ValWood:
yes you are right @jseager7. Showing only the genes for the organism being annotated would fix this issue.

jseager7:
@kimrutherford It sounds like this issue could require changes to the annotation extension configuration syntax (unless we can handle this with further processing on the front-end).

Currently, if you use the 'GeneID' range on an extension, Canto shows all the genes in the session (for every organism), but it sounds like we want to only see the genes for the 'current' organism (that is, the one being annotated with a phenotype term).

I think we either need:

[...]

PR #2111, opened 2019-11-13
Add organism name after genes in feature chooser

jseager7:
(Temporary fix for #2101)

This pull request appends organism names to genes in the feature chooser when in multi-organism mode (and pathogen-host mode). That makes it possible to tell the difference between two or more genes with the same name from different organisms.

This is intended to be a temporary fix until we can implement organism selectors into the workflow so that the gene lists can always be filtered by organism.

Here's how it looks:

image

- - -

@kimrutherford I'm not able to test this on single organism mode because my database fails when trying to upgrade to schema version 30 or 31 (not sure which). The error's below:

```
upgrading to version 30
DBIx::Class::Storage::DBI::_prepare_sth(): DBI Exception: DBD::SQLite::db 
prepare_cached failed: no such table: allele_note [for Statement "DELETE FROM 
allele_note WHERE ( allele_note_id IN ( SELECT allele_notes.allele_note_id 
FROM allele me  JOIN allele_note allele_notes ON allele_notes.allele = 
me.allele_id WHERE ( allele_id NOT IN (SELECT allele FROM allele_genotype) ) 
) )"] at lib/Canto/Curs/GenotypeManager.pm line 232
```
In the meantime, do you think you could test this change to make sure the gene names don't appear in the drop-down in single organism mode?

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -380,10 +380,12 @@ function symbolEncoder() {
 canto.filter('encodeAlleleSymbols', symbolEncoder);
 canto.filter('encodeGeneSymbols', symbolEncoder);
 
-
-canto.filter('featureChooserFilter', function () {
+canto.filter('featureChooserFilter', ['CantoGlobals', function (CantoGlobals) {
   return function (feature) {
     var ret = feature.display_name;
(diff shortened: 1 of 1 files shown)


kimrutherford:
Thanks James.  I tested in pombe mode and in PHI mode and it works for me.

PR #2114, opened 2019-11-15
Always show autocomplete for short lists

### LINKED ISSUE #2038: include definitions for "extent of infectivity"
I find it hard to differentiate these but I think we have definitions for them somewhere?
Can these be included ?

defs

kimrutherford:
Hi @jseager7 

This PR attempts to fix #2038.  Could you test it out at some point?  It's low priority though.


kimrutherford:
Thanks for testing.  I'll add a tooltip to the OK button before merging.

kimrutherford:
> I'll add a tooltip to the OK button before merging.

Done!

Issue #2115, opened 2019-11-18
Restore gene-gene physical interactions


kimrutherford:
See: https://github.com/pombase/canto/issues/2080#issuecomment-554122662

kimrutherford:
This is nearly fixed now.  I think the changes should be ready for testing tomorrow.

kimrutherford:
@jseager7 does your main Canto have any physical interactions?  This command line will find them:
```
./canto/script/canto_docker ./script/canto_curs_map.pl 'my $rs = $curs_schema->resultset("Annotation")->search({ type => "physical_interaction" }); print $curs_key, " ", $rs->count(), "\n" if $rs->count() > 0;'
```

kimrutherford:
I've merged the changes.  I'll close this.  We can open new issues for problems.

PR #2116, opened 2019-11-19
Restore gene-gene interactions

kimrutherford:
This PR contains the changes to restore gene-gene interaction annotation so we can handle physical interactions.

This is a work in progress.

See #2115.


kimrutherford:
Hi James.

This is a rather big change so could I ask you to give it a test?  I think everything is done now.
I'll be doing more testing myself tomorrow.

kimrutherford:
Hi James.  Thanks for the thorough testing.

> When you edit an existing Physical Interaction, the 'Organism' field doesn't default to the organism
> that was already set

I think I've fixed that problem now.

> Check for duplicate physical interactions

I made a separate issue for that: https://github.com/pombase/canto/issues/2123

> Exclude wild-type host organisms (with no genes)

New issue: https://github.com/pombase/canto/issues/2124

> Genes can interact with themselves

That one is OK.

> Interacting gene goes unset when finishing interaction

That's fixed now.

> Handle disabled fields differently
> Curator field is blank when creating an interaction

I've removed the curator field.  We should have a chat on one of the Monday calls about what to do about the other headings for disabled fields.

Cheers!

jseager7:
> We should have a chat on one of the Monday calls about what to do about the other headings for disabled fields.

I've opened an issue for this in the meantime at https://github.com/pombase/canto/issues/2126.

Issue #2128, opened 2019-11-26
In annotation dialog, don't allow change of organism if the gene can't change


kimrutherford:
If we go to the annotation dialog using quick add on a gene page, we shouldn't be showing a drop-down for the organism in the annotation:

![annotation-edit-gene-org-1](https://user-images.githubusercontent.com/90474/69589401-da0dff80-1050-11ea-89e0-468c15fa4190.png)

kimrutherford:
I think I've fix this.  Please let me know if you see any problems.

PR #2131, opened 2019-11-26
Move Physical Interaction link to gene page

### LINKED ISSUE #2108: Physical interaction quick link always visible
I've noticed in the latest code that the Quick links section is always visible, whether you're in advanced mode or not, because the 'physical interaction' link is always visible:

image

 I'm assuming this is because there's no other place in the interface to access 'physical interaction' now? I can't see it in the list of annotation types on the gene page.

jseager7:
Fixes #2108 

This pull request moves the Physical Interaction link back to the gene page, so that it doesn't need to be permanently visible on the curation summary page. I think this makes sense now that physical interactions use genes again (instead of genotypes).

Note that the Physical Interaction link always uses the 'Quick Add' workflow style (the modal-based workflow) when on the gene page, unlike the other annotation types that have the option for a more guided workflow. As far as I know there isn't any guided workflow for physical interaction annotations.

### DIFF
--- a/root/curs/front_page_quick_links.mhtml
+++ b/root/curs/front_page_quick_links.mhtml
@@ -3,14 +3,14 @@
 </%args>
 
 % if (!$read_only_curs) {
-<div class="curs-box">
+<div ng-if="getAnnotationMode() == 'advanced'" class="curs-box">
   <div class="curs-box-title">
     Quick links
   </div>
   <div class="curs-box-body">
     <ul class="curs-quick-links-list">
 % for my $annotation_type (@annotation_type_list) {
-    <li ng-if="getAnnotationMode() == 'advanced' || '<% $annotation_type->{category} %>' == 'interaction'">
+    <li>
       <annotation-quick-add annotation-type-name="<% $annotation_type->{name} %>"
                             link-label="<% ucfirst($annotation_type->{display_name}) %>"
                             feature-type="<% $annotation_type->{feature_type} %>"
--- a/root/curs/gene_page.mhtml
+++ b/root/curs/gene_page.mhtml
@@ -23,14 +23,14 @@ Choose curation type for <% $gene->display_name() %>:
 %   my $type_display_name = $annotation_type->{display_name};
 %   if ($annotation_type->{feature_type} eq 'gene') {
 %    if ($annotation_type->{category} eq 'interaction') {
-%#    <li class="annotation-type">
-%#      <annotation-quick-add annotation-type-name="<% $annotation_type->{name} %>"
-%#                            feature-type="gene" feature-id="<% $gene->gene_id() %>"
-%#                            feature-display-name="<% $gene_proxy->display_name() %>"
-%#                            link-label="<% ucfirst $type_display_name %>">
-%#      </annotation-quick-add>
-%#      <help-icon key="<% $annotation_type->{name} . '_definition' %>"></help-icon>
-%#    </li>
+    <li class="annotation-type">
+      <annotation-quick-add annotation-type-name="<% $annotation_type->{name} %>"
+                            link-label="<% ucfirst($annotation_type->{display_name}) %>"
+                            feature-type="<% $annotation_type->{feature_type} %>"
+                            feature-taxon-id="">
+      </annotation-quick-add>
+      <help-icon key="<% $annotation_type->{name} . '_definition' %>"></help-icon>
+    </li>
 %    } else {
 %     my $action_path = "$curs_root_uri/feature/gene/annotate/$gene_id/" .
 %       "start/$annotation_type_name";


root/curs/front_page_quick_links.mhtml
```diff
@@ -3,14 +3,14 @@
 </%args>
 
 % if (!$read_only_curs) {
-<div class="curs-box">
```
kimrutherford: I'm not sure about this change.  Doesn't it mean that you can't make a "Genetic interaction" annotation unless you're on advanced mode?

root/curs/gene_page.mhtml
```diff
@@ -23,14 +23,14 @@ Choose curation type for <% $gene->display_name() %>:
...
-%#      <annotation-quick-add annotation-type-name="<% $annotation_type->{name} %>"
-%#                            feature-type="gene" feature-id="<% $gene->gene_id() %>"
-%#                            feature-display-name="<% $gene_proxy->display_name() %>"
-%#                            link-label="<% ucfirst $type_display_name %>">
-%#      </annotation-quick-add>
-%#      <help-icon key="<% $annotation_type->{name} . '_definition' %>"></help-icon>
-%#    </li>
+    <li class="annotation-type">
+      <annotation-quick-add annotation-type-name="<% $annotation_type->{name} %>"
+                            link-label="<% ucfirst($annotation_type->{display_name}) %>"
```
kimrutherford: It would be good to use the feature-id and feature-taxon-id attributes on annotation-quick-add so that the gene and organisms are preset in the dialog:

```
     <li class="annotation-type">
       <annotation-quick-add annotation-type-name="<% $annotation_type->{name} %>"
                             link-label="<% ucfirst($annotation_type->{display_name}) %>"
+                            feature-id="<% $gene->gene_id() %>"
                             feature-type="<% $annotation_type->{feature_type} %>"
-                            feature-taxon-id="">
+                            feature-display-name="<% $gene_proxy->display_name() %>"
+                            feature-taxon-id="<% $gene_proxy->taxonid() %>">
       </annotation-quick-add>
       <help-icon key="<% $annotation_type->{name} . '_definition' %>"></help-icon>
     </li>
```

kimrutherford:
Thanks James.  I've added two comments.

root/curs/front_page_quick_links.mhtml
jseager7: You're right; I misread line 13 as referring to only physical interaction annotations, rather than all annotations in the interaction category:

```html
<li ng-if="getAnnotationMode() == 'advanced' || '<% $annotation_type->{category} %>' == 'interaction'">
```
I didn't notice any problems at the time because PHI-Canto doesn't have genetic interaction enabled. I realise now that genetic interaction only applies to metagenotypes, so there isn't really any other place for it at the moment. I guess the only other place would be to add a 'Annotate metagenotypes' heading under 'Annotate genotypes' and put it there, along with the Metagenotype Management link (since that isn't strictly about annotating genotypes either.

For now, can we change this check to apply only to genetic interactions?

root/curs/gene_page.mhtml
jseager7: Thanks. When I made this change I just copied the code from the summary page because I wasn't sure what these other attributes did or how they should be used. I'll test this change now.

jseager7: I've tested now and there's multiple problems with setting these attributes: firstly, the physical interaction annotation doesn't perform the filtering by organism role that we have in #2132, so even though the organism and the gene sets correctly, there's nothing to stop you switching to another host or pathogen. On top of that, you can't edit the gene after it's first set, so you can end up with invalid organism / gene combinations like this:

image

(where TRI5 is a _F. graminearum_ gene).

Secondly, I've noticed that the 'Interacting gene' field never seems to enable after I added the extra attributes. It stays blank after picking an interaction type, even if the gene / organism combination is valid:

image

I think this could be to do with the fact the gene isn't selected by the user.

kimrutherford: > so even though the organism and the gene sets correctly, there's nothing to stop you switching to another host or pathogen

I think that's this issue: #2128 
I plan to work on that soon.

> It stays blank after picking an interaction type

I'm off to bed now but I'll dig into the logic tomorrow to find a fix.

root/curs/front_page_quick_links.mhtml
kimrutherford: > For now, can we change this check to apply only to genetic interactions?

I've done that tweak.

kimrutherford:
> Secondly, I've noticed that the 'Interacting gene' field never seems to enable after I added the
> extra attributes

I've merged your pull request and checked in a fix for this problem.

kimrutherford:
Please let me know if you spot any problems after these changes.

PR #2132, opened 2019-11-26
Filter organisms by role in annotation dialog

### LINKED ISSUE #2122: Restrict organism to host or pathogen in annotation dialog
Currently if you open an annotation dialog from the "Host phenotype" or "Pathogen phenotype" quick links, it shows all genotypes in both cases.  We should restrict the drop-down to show host or phenotype genotypes only (as appropriate).

jseager7:
Fixes #2122 

This is a quick fix to filter the list of organisms by their pathogen/host role when creating a 'pathogen phenotype' or 'host phenotype' annotation using the Quick Links.

I haven't thoroughly tested this in other modes, but I think the conditions are so restrictive (exact matching on the annotation type name) that it shouldn't affect anything except PHI-Canto.

It would be nice if we had a more flexible way of testing whether an annotation type applied to pathogens or hosts (perhaps a config option analogous to the feature type), but we probably don't need to worry about that yet.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -6371,13 +6371,13 @@ var annotationEditDialogCtrl =
 
     $scope.filteredOrganismPromise
       .then(function (organisms) {
-        if (organisms.length == 1) {
-          $scope.selectedOrganism = organisms[0];
-        }
         if (args.annotationTypeName === 'host_phenotype') {
           organisms = filterOrganisms(organisms, 'host');
         } else if (args.annotationTypeName === 'pathogen_phenotype') {
           organisms = filterOrganisms(organisms, 'pathogen');
+        }
+        if (organisms.length == 1) {
+          $scope.selectedOrganism = organisms[0];
         }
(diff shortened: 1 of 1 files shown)


root/static/js/canto-modules.js
```diff
@@ -6374,6 +6374,11 @@ var annotationEditDialogCtrl =
         if (organisms.length == 1) {
```
kimrutherford: Perhaps the check to see if there is only one organism should go after the filter for host or pathogen?

jseager7: Yes, that makes more sense. I'll push a change for that.

jseager7:
I've moved the default organism check to after the organism-role filtering. I've also made the organism length check use strict equality, since I don't think that `.length` can be non-numeric (or not anything that would fit loose equality like strings, anyway):

```js
if (organisms.length === 1) {
```
Please let me know if it was using loose equality for any special reason.

kimrutherford:
Thanks James.  That all looks great now.

PR #2133, opened 2019-11-26
Handle read-only sessions in genotypeAndSummaryNavCtrl

### LINKED ISSUE #2074: New navigation links don't respect read-only mode
I've added multiple new navigation links to Canto – most of which affect PHI-Canto but some affect other instances too – but I forgot to handle the case where the session is in read-only mode, meaning these links have the effect of putting any viewer back into edit mode.

jseager7:
Fixes #2074 

This pull request fixes an oversight in `genotypeAndSummaryNavCtrl` where the directive didn't account for the session being in read-only mode. I've tested this in single organism mode and there's no problems there.


kimrutherford:
Thanks, looks good to me.

PR #2149, opened 2019-12-04
Disable host genotype link until hosts are added

### LINKED ISSUE #2148: Disable host genotype management link when there are no hosts
PHI-Canto allows users to start a curation session with only pathogens in the session. This is intended, but Canto doesn't disable the link to the Host genotype management page when there are no hosts. We already disable the Metagenotype management link when a metagenotype can't be created, so we should do the same thing here.

jseager7:
Fixes #2148 

This PR disables the Host genotype management link on the summary page when there are no hosts in the session.

I changed the `_metagenotype_flags` subroutine in `lib/Canto/Controller/Curs.pm` so that it also returns the `$has_host` variable (rather than only using the variable to test if the metagenotype link should be enabled). That means that the subroutine isn't really accurately named anymore, since it doesn't relate only to metagenotypes. I don't know whether it would be better to rename the subroutine, or to extract the genotype link flag setting into a different subroutine.

Also, I didn't update the tests for this module, because I wasn't sure where to find them (or if there even are any tests for this functionality). I'm just going to open the PR and see what Travis-CI says.

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -174,7 +174,7 @@ sub top : Chained('/') PathPart('curs') CaptureArgs(1)
   $st->{edit_organism_page_valid} = 0;
 
   if ($st->{pathogen_host_mode}) {
-    ($st->{show_metagenotype_links}, $st->{edit_organism_page_valid}) =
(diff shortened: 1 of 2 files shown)


kimrutherford:
That's great.  The code looks correct and it worked when tested.  Thanks!

Issue #2125, opened 2019-11-25
Disable phenotype quick links if there are no genotypes


kimrutherford:
In PHI mode disable host phenotype quick link if there are no host genotypes.  Similarly for pathogens.

kimrutherford:
@jseager7 I've just created a pull request that hopefully fixes this.  At the same time I've added code to disable the "Pathogen genotype management" link on the front page if there are no genes.

Could you give this change a test when you have a chance?  Please go ahead and merge it if it looks OK, otherwise I'll fix the problems you find.  :-)

jseager7:
Closed by #2152.

Issue #2102, opened 2019-11-07
Multiple ways for users to avoid strain selection


jseager7:
## Method 1: Exiting and re-entering the session

Canto's curation state management system doesn't check whether the user needs to enter strains for their organisms when a user returns to a session after exiting it.

That means if a user navigates away from Canto when they're on the confirmation page (`edit_genes`), upon returning to the session they will be put onto the curation summary page without having added any strains. Then, when they try to create a genotype, the strain selection only shows the placeholder option 'Choose a strain...', but of course they can't do that because there are no strains.

![image](https://user-images.githubusercontent.com/37659591/68393712-7b95e580-0164-11ea-828c-aaf83c75fa06.png)

## Method 2: Breadcrumb links

It's also possible to escape the strain selection page by clicking the summary link in the breadcrumb trail, because the link isn't disabled. It would probably be better to just hide the breadcrumb trail here, or at least make it only link back to the gene entry page, and not forward to the summary page.

![image](https://user-images.githubusercontent.com/37659591/68394150-3920d880-0165-11ea-9a5c-857fb32d943c.png)

- - -

I'm assigning this to @kimrutherford because it probably involves back-end routing work, but I'm happy to take a look at any front-end integration.

kimrutherford:
I've had a go at fixing this.  The user now gets sent to the gene edit page if there are missing strains.

jseager7:
@kimrutherford I've tested the bug about re-entering the session, and it looks like that's fixed, thanks. However, it's still possible to escape the organism confirmation page by clicking the summary page link in the breadcrumb trail. Do we want to do anything about this?

kimrutherford:
> However, it's still possible to escape the organism confirmation page by clicking the summary
> page link in the breadcrumb trail. 

Hi James.  I can't reproduce that.  For me it always goes back to the gene and organisms confirmation page.  Let's chat about that on Skype.

jseager7:
I must not have actually tested what the link did, because you're right, it does redirect to the organism confirmation. Closing again.

PR #2152, opened 2019-12-09
Disable links when genes/genotypes are missing

kimrutherford:
Disable host/pathogen phenotype Quick link when host/pathogen genotypes are
missing.  And disable pathogen gentype management link when there are no
pathogen genes.

Refs pombase/canto#2125

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -174,7 +174,10 @@ sub top : Chained('/') PathPart('curs') CaptureArgs(1)
   $st->{edit_organism_page_valid} = 0;
 
   if ($st->{pathogen_host_mode}) {
-    ($st->{show_metagenotype_links}, $st->{show_host_genotype_link}, $st->{edit_organism_page_valid}) =
+    ($st->{show_metagenotype_links}, $st->{show_host_genotype_link},
+     $st->{show_pathogen_genotype_link},
+     $st->{has_host_genotypes}, $st->{has_pathogen_genotypes},
+     $st->{edit_organism_page_valid}) =
       _metagenotype_flags($config, $schema);
   }
 
@@ -342,6 +345,8 @@ sub _metagenotype_flags
   my $rs = $schema->resultset('Organism');
 
   my $has_host = 0;
+  my $has_pathogen_genes = 0;
+  my $has_host_genotypes = 0;
   my $has_pathogen_genotypes = 0;
 
   my $organism_page_valid = 1;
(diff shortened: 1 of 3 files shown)


lib/Canto/Controller/Curs.pm
```diff
@@ -342,6 +345,8 @@ sub _metagenotype_flags
   my $rs = $schema->resultset('Organism');
 
   my $has_host = 0;
+  my $has_pathogen_genes = 0;
```
jseager7: Strictly speaking, checking for pathogen genes shouldn't be necessary, since PHI-Canto forbids starting a session if you don't have any pathogen genes. It allows a session with no hosts, but not the contrary.

I first thought that the only reason to handle the lack of pathogens genes is if it were possible (or will be possible later) to remove the pathogen genes when you're not on the confirmation page &ndash; however, it turns out the fix for escaping the organism confirmation page (#2102) doesn't cover the case where the user hasn't added pathogen genes, so it's still possible to end up on the summary page with no pathogen genes, which makes this fix necessary (although we should really have a fix for the case I've described too).

Given that you've already added the logic, it's probably not worth taking it out, since it does future-proof us against workflow bugs, plus any future need to allow sessions without pathogen genes.

root/curs/front_page_quick_links.mhtml
```diff
@@ -10,13 +10,22 @@
   <div class="curs-box-body">
     <ul class="curs-quick-links-list">
 % for my $annotation_type (@annotation_type_list) {
+%   my $feature_subtype = $annotation_type->{feature_subtype} || 'NONE';
+%   my $show_link = ($feature_subtype eq 'NONE' || $feature_subtype eq 'host' && $has_host_genotypes || $feature_subtype eq 'pathogen' && $has_pathogen_genotypes) ? 'true' : 'false';
```
jseager7: The handling for the 'Pathogen-host interaction phenotype link' is inconsistent with the logic that we use here, because for metagenotype annotations, Canto treats the underlying concept (the metagenotype) as as interaction (like a genetic interaction). In this case, the link is always enabled, but you get a warning pop-up when clicking the link:

image


image

This message is a bit too generic for pathogen-host interactions, since the minimum required to create a metagenotype is _specifically_ one pathogen genotype (you can't create one if you just add host genotypes). Note that you don't need host genotypes because you can always use the default wild type host genotype on the Metagenotype Management page.

kimrutherford:
Hi James.  Thanks for testing.

> Given that you've already added the logic, it's probably not worth taking it out, since it does
> future-proof us against workflow bugs

That was my thinking too.  And I thought I was more likely to break things if I tried to remove that logic.

> This message is a bit too generic for pathogen-host interactions, since the minimum
> required to create a metagenotype is specifically one pathogen genotype

Could you make a issue about that?

jseager7:
> Could you make a issue about that?

This is now open at #2173.

PR #2155, opened 2019-12-10
Disable host genotype link until host gene exists

### LINKED ISSUE #2059: Improve message on management page for hosts with no genotypes
Currently it says "No genotypes in this session. Try adding one." for hosts with no genes but there is no way to add a genotype:

canto-host-no-genotypes-1

jseager7:
Fixes #2059 

This pull request amends the condition that disables the Host genotype management link, so that it disables when there are no host genes, rather than no host organisms. This is to accommodate the case where a host has been added without genes (which is only possible for host organisms, not pathogens). It doesn't make sense to give access to the Host genotype management link when a genotype can't be created.

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -346,6 +346,7 @@ sub _metagenotype_flags
 
   my $has_host = 0;
   my $has_pathogen_genes = 0;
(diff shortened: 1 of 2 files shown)


kimrutherford:
Looks good.  Thanks James.

Issue #1846, opened 2019-04-30
Experimental conditions autocomplete doesn't behave consistently


jseager7:
(Requested by @CuzickA)

The conditions picker doesn't always work in a consistent way, such that:

- keyboard presses only register under certain conditions, and
- text that the user enters isn't always converted to a tag.

Here's a list of the specific issues:

## Text doesn't convert to tag

This bug can be triggered by first typing a fragment of a term that triggers the autocomplete drop-down (for example, 'high'):

![condition_autocomplete](https://user-images.githubusercontent.com/37659591/56963569-2ba0eb00-6b51-11e9-834d-6e274b28facf.png)

Then, if you lose focus on the text input box (by clicking elsewhere), the text doesn't automatically get converted into a tag:

![autocomplete_no_tag_convert](https://user-images.githubusercontent.com/37659591/56963570-2e034500-6b51-11e9-90a2-7edd1aa329e9.png)

This means that the condition won't actually be saved when you close the popup, which also has the effect of preventing the user from 'finishing' the edit if they haven't changed any other fields &ndash; even though it looks like they have.

The text still converts to a tag when you regain focus (click back onto) the text input box.

**Note** that this automatic conversion on loss of focus (the expected behaviour) _does_ happen if you enter any text that _doesn't_ trigger the autocomplete drop-down (meaning, any term that isn't in the condition list, such as nonsense like 'xxyyzz').

## Keyboard actions are sometimes ignored

Normally, when you press <kbd>Tab</kbd> or <kbd>Enter</kbd> when the drop-down _isn't_ open, the text gets converted to a tag:

![autocomplete_tab_action](https://user-images.githubusercontent.com/37659591/56964062-50e22900-6b52-11e9-9e87-ac40c333a269.png)

However, when the autocomplete is open, <kbd>Tab</kbd> and <kbd>Enter</kbd> don't do anything.

I think that <kbd>Tab</kbd> should focus the first entry on the drop-down list, like what currently happens when you hit the down arrow key <kbd>&darr;</kbd>:

![autocomplete_focus](https://user-images.githubusercontent.com/37659591/56964353-ef6e8a00-6b52-11e9-93da-038dea9df177.png)

An alternative (probably more conventional) is that <kbd>Tab</kbd> should focus the next input on the page; currently, it doesn't even do that.

As for what <kbd>Enter</kbd> should do, that's debatable. I think there are two possible options:

[...]

PR #2158, opened 2019-12-11
Allow autocomplete when suggestion list is open

### LINKED ISSUE #2025: Can't complete custom experimental conditions that partially match
(Sub-issue of #1846)

**Summary:** <kbd>Enter</kbd> and <kbd>Tab</kbd> don't autocomplete experimental conditions until something is selected from the list; the shortcuts also don't autocomplete custom entries (unless there are zero matches in the drop-down). We should make <kbd>Enter</kbd> and <kbd>Tab</kbd> complete the current entry unless a drop-down entry is selected.

### Problem

@CuzickA has noticed that if you type an experimental condition that isn't in the autocomplete list, but still partially matches existing terms &ndash; for example 'host kill assay' &ndash; the autocomplete doesn't allow you to complete your custom term with <kbd>Enter</kbd> or <kbd>Tab</kbd>.

The image below shows an example of when neither <kbd>Enter</kbd> nor <kbd>Tab</kbd> do anything:

image

This is annoying, since the only way to force the term to complete is to break focus (click out) on the form and then re-enter it. If the user clicks out but _doesn't_ re-enter the form, the term doesn't even complete and the user won't be able to finish editing the annotation.

While testing this problem, I noticed that <kbd>Enter</kbd> and <kbd>Tab</kbd> don't seem to do anything until you select an entry from the drop-down. I expected it to autocomplete by choosing the first item in the list in this case.

For example, in the image below, <kbd>Enter</kbd> and <kbd>Tab</kbd> still don't autocomplete '+&nbsp;triclosan', even though there is only one suggestion:

image

### Solution

Seeing as <kbd>Enter</kbd> and <kbd>Tab</kbd> have no defined functionality except when a term is selected, I think the best solution is to **make <kbd>Enter</kbd> and <kbd>Tab</kbd> finish the entry that the user has typed if they haven't selected anything from the drop-down;** otherwise, trigger the autocomplete (if they have selected something).

Note that this solution basically precludes using <kbd>Enter</kbd> and <kbd>Tab</kbd> to autocomplete the first entry in the list on a partial match. We could choose to make these shortcuts do different things in order to handle both cases, but we'd likely have to explain this to users.

jseager7:
(Fixes #2025, partial fix for #1846)

This is a small patch copied from the tag-it repository, which allows tags to be created when the autocomplete menu is still open.

@kimrutherford We could also apply this by updating tag-it to the latest source version, but I seem to remember you tried this before and it broke the widget. This patch seems to work fine, though. Let me know if you want me to try updating the whole file instead, since that might help fix the other problem in issue #1846.

### DIFF
--- a/root/static/js/tag-it.js
+++ b/root/static/js/tag-it.js
@@ -266,7 +266,8 @@
                         }
 
                         // Autocomplete will create its own tag from a selection and close automatically.
-                        if (!that.tagInput.data('autocomplete-open')) {
+                        if (!(that.options.autocomplete.autoFocus && that.tagInput.data('autocomplete-open'))) {
+                            that.tagInput.autocomplete('close');
                             that.createTag(that._cleanedInput());
                         }
                     }


kimrutherford:
Excellent stuff.  I've tested this and it's a big improvement.  Cheers!

> We could also apply this by updating tag-it to the latest source version

I think we should merge this fix for now and try the latest version at some point.  I made an issue for that: #2159

PR #2162, opened 2019-12-12
Fix experimental conditions not adding on blur

### LINKED ISSUE #2159: Try the latest tagit.js version
(from pombase/canto#2158)

We should try the most recent version of tagit.js and use it if possible.

https://github.com/aehlke/tag-it

jseager7:
Fixes #2162 

This is a one-line fix copied from an unmerged pull request (`https://github.com/aehlke/tag-it/pull/321`) on the tag-it repository. It ensures that any typed tags are always finished (added) when the user breaks focus on the widget's text field. I've tested the fix and haven't noticed any problems with it.

@kimrutherford I'm leaving this just for you to double-check, but feel free to squash-merge this if you don't notice any problems.

### DIFF
--- a/root/static/js/tag-it.js
+++ b/root/static/js/tag-it.js
@@ -274,7 +274,7 @@
                 }).blur(function(e){
                     // Create a tag when the element loses focus.
                     // If autocomplete is enabled and suggestion was clicked, don't add it.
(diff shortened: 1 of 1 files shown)


kimrutherford:
Thanks for the fix.

PR #2164, opened 2019-12-12
Update tag-it source files to latest version

jseager7:
References #2159

This pull request updates the JS and CSS files for tag-it to the latest commit on the tag-it repository (even though that was six years ago).

@kimrutherford I've tested the latest version of the tag-it widget and I can't notice any problems. We can probably merge this and see what happens, but note that there's a possibility this PR could conflict with commit 2f19a53a18d4e1dc561eb07c81c1d1054a7dc06d and PR #2162. Both of those changes contain fixes that are more important than this PR. I haven't had a chance to test this PR with those newer changes merged in yet.

### DIFF
--- a/root/static/css/jquery.tagit.css
+++ b/root/static/css/jquery.tagit.css
@@ -1,54 +1,69 @@
 ul.tagit {
-  padding: 1px 5px;
-  overflow: auto;
-  margin-left: inherit; /* usually we don't want the regular ul margins. */
-  margin-right: inherit;
+    padding: 1px 5px;
+    overflow: auto;
+    margin-left: inherit; /* usually we don't want the regular ul margins. */
+    margin-right: inherit;
 }
 ul.tagit li {
-  display: block;
-  float: left;
-  margin: 2px 5px 2px 0;
+    display: block;
+    float: left;
+    margin: 2px 5px 2px 0;
 }
-ul.tagit li.tagit-choice {
-  padding: .2em 18px .2em .5em;
-  position: relative;
-  line-height: inherit;
+ul.tagit li.tagit-choice {    
+    position: relative;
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks for giving that a try.  I'll test it when I'm back from my holidays.

kimrutherford:
Hi James.  I've finally tried this PR.  Sorry it took so long.  It all works for me so I think we should merge it.  I tested by merging locally into the master branch.  There are no conflicts.

One thing though, could you tell me about the "destroy" function you've added?  Is it something needed when using the new tag-it version?

jseager7:
> One thing though, could you tell me about the "destroy" function you've added? Is it something needed when using the new tag-it version?

It's been so long I can barely remember, but based on my original comment, it looks like all the changes are just a result of updating the tag-it source files to match what's in the tag-it repository. So it wasn't really me who added the change &ndash; I just pulled the latest changes from another repository.

For what it's worth, here's the commit that adds the `destroy` function in the upstream repository: https://github.com/aehlke/tag-it/commit/bd5bb36e1f86a6def743ec4b88d2ee7a11ea9373

kimrutherford:
> So it wasn't really me who added the change 

Ah yes!  Sorry about that.  I'll merge this now.

jseager7:
I've tested this against the patches applied in https://github.com/pombase/canto/pull/2158/commits/24ae1236875764b44409063ee78fb30719d8f11b and https://github.com/pombase/canto/pull/2162/commits/57bbe92df2224e91675f9ecc1539f59610cabf02, and there don't seem to be any problems introduced by this update (I think the patch in 24ae1236875764b44409063ee78fb30719d8f11b was already merged into the latest version of the tag-it.js repository anyway).

PR #2175, opened 2020-01-08
Merge canto_start_docker scripts and add "--auto"

kimrutherford:
Refs PHI-base/config#24


script/canto_start_docker
```diff
@@ -23,12 +34,21 @@ then
...
-if [[ "$1" == "--debug" ]]; then
+if [ "$1" == "--debug" ]; then
+    shift
     DEBUG="-e CANTO_DEBUG=TRUE"
 fi
 
-CANTO_DOCKER_RUN_ARGS="$EXTRA_ARGS --rm -ti $DEBUG --security-opt seccomp=unconfined --cap-add=SYS_PTRACE --mount type=bind,source=$(pwd)/data,target=/data --mount type=bind,source=$(pwd)/import_export,target=/import_export --mount type=bind,source=$(pwd)/canto,target=/canto -w=/canto pombase/canto-run:v8"
+TTY_OPT="-t"
+
+if [ "$1" == "--auto" ]; then
```
jseager7: With regards to your comment about the naming of the `--auto` flag, would it be better to base it on Docker's naming convention for the flag? They use `--tty` as the long name for the `-t` flag, so maybe we could call our flag `--no-tty`? You're already naming the variable according to its usage for controlling the TTY, after all.

kimrutherford: Good suggestion.  Thanks, I've done that.

Issue #2156, opened 2019-12-10
Use organism names in metagenotype display name


jseager7:
@CuzickA has noticed that the current metagenotype display name isn't useful when you have a wild-type host with no genes. All you get is the text 'host: wild type', which tells you nothing about what the host species is:

![image](https://user-images.githubusercontent.com/37659591/70537700-d0a88b00-1b58-11ea-97db-c717b60b05ff.png)

This display name is also inconsistent with how we display the metagenotype name in the annotation edit modal:

![image](https://user-images.githubusercontent.com/37659591/70537800-f7ff5800-1b58-11ea-958d-89b8cb8d2593.png)

...and in the annotation tables on the summary page:

![image](https://user-images.githubusercontent.com/37659591/70537911-211fe880-1b59-11ea-8ed9-449f13fee0b8.png)

We should amend the default display name to include more information (at least the organism name, though we probably need the strain name too).

jseager7:
I've looked into this, but I'm not sure there's much I can do because the `display_name` is generated on the back-end.

Specifically, it looks like `lib/Canto/CursDB/Metagenotype.pm` doesn't have any way to get the organism name or the strain name. For example, you can't do `$self->pathogen_genotype->organism->full_name()`. This isn't the same as using the `Metagenotype` service on the front-end; that has all of the organisms and strain names joined to the host and pathogen genotypes.

@kimrutherford how should we proceed? Should the back-end include organism and strain information when returning a metagenotype, or should we try to migrate more of these templates (`title.mhtml`, `metagenotype_page.mhtml`, and so on) to AngularJS, so we can use the AngularJS services instead?

kimrutherford:
> Should the back-end include organism and strain information when returning a metagenotype,
> or should we try to migrate more of these templates (title.mhtml, metagenotype_page.mhtml,
> and so on) to AngularJS, so we can use the AngularJS services instead?

I think both of those make sense but probably returning more information from the back-end is the quicker fix.  Let's chat about this next time on Skype.

kimrutherford:
> I think both of those make sense but probably returning more information from the back-end
> is the quicker fix.

Now I've checked, I think this information is already available.  The metagenotype list service returns separate "host_genotype" and "pathogen_genotype" fields which could be used to create a display name in the JS code.  Here's a truncated JSON example from the service:

```json
[
  {
    "feature_id": 1,
[...]

PR #2193, opened 2020-01-24
Include the organism and strain in the metagenotype display name

kimrutherford:
Refs #2156

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -1648,7 +1648,7 @@ sub feature_view : Chained('feature') PathPart('view')
         $st->{features} = [$metagenotype];
 
         my $display_name = $st->{feature}->display_name($config);
-        $st->{title} = "Metagenotype: $display_name";
+        $st->{title} = "Metagenotype details";
 
       } else {
         die "no such feature type: $feature_type\n";
--- a/lib/Canto/CursDB/Metagenotype.pm
+++ b/lib/Canto/CursDB/Metagenotype.pm
@@ -220,13 +220,63 @@ sub all_annotations
   return $self->annotations();
 }
 
+use feature "state";
+
+use Canto::Track::AdaptorUtil;
+
 sub display_name
 {
   my $self = shift;
   my $config = shift;
 
-  return 'pathogen: ' .$self->pathogen_genotype->display_name($config) . ' / host: ' .
-    $self->host_genotype->display_name($config);
+  state $organism_lookup = Canto::Track::AdaptorUtil::get_adaptor($config, 'organism');
+  state $strain_lookup = Canto::Track::AdaptorUtil::get_adaptor($config, 'strain');
+
+  my $first_genotype = $self->first_genotype();
+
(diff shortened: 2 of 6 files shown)


jseager7:
Also, if FlyBase uses metagenotypes, how useful is it to show the organism names in their case? Presumably this applies to all forms of metagenotypes, not just pathogen-host interactions.

ValWood:
Just a thought but would it make sense to make this header text appear below the breadcrumb left justified instead of  to the right of the breadcrumb?

jseager7:
> Just a thought but would it make sense to make this header text appear below the breadcrumb left justified instead of to the right of the breadcrumb?

That's an option, although there's a Description section that's not shown in the screenshots above that contains the same information, so having the header text at all seems slightly redundant:

image

So another option is to break that Description section up to describe the constituent parts of the metagenotype, then have a subheading with the metagenotype display name, and finally a generic title to keep the page layout consistent with the rest of Canto.

image

This also has the advantage that we can link to the genotype pages for the constituent genotypes, and italicise the species names.

jseager7:
Although note that we'd probably want to use different terminology from 'pathogen' and 'host' when describing interactions (metagenotypes) that aren't pathogen-host interactions.

ValWood:
I think the reason we haven't done this is that the header is consistent on every page  (genotype/ gene PMID etc),  Although this could still work @kimrutherford 


> Although note that we'd probably want to use different terminology from 'pathogen' and 'host' when describing interactions (metagenotypes) that aren't pathogen-host interactions.

Aren't all  metagenotypes pathogen host interactions though?

jseager7:
> Aren't all metagenotypes pathogen host interactions though?

Perhaps not anymore, since I think the metagenotype concept has been generalised in Canto to cover any multi-organism interaction. This allows the metagenotype to be used as the underlying concept (annotation target) for physical interactions and genetic interactions. In that sense, the pathogen-host interaction is a specialisation of a metagenotype.

So, assuming FlyBase is using physical interactions, genetic interactions, and/or annotations made through the metagenotype management page, any display changes that apply to metagenotypes will have knock-on effects for FlyBase. Having said that, I don't know whether or not FlyBase use the pages that will be affected by this specific change.

If this page was used by FlyBase, they'd probably prefer to see 'Interactor A' and 'Interactor B' instead of 'pathogen' or 'host'. That's the convention followed in the interface for Physical Interactions and Genetic Interactions:

image

ValWood:
Oh I see, pathogen and host would not work here.

kimrutherford:
> Although note that we'd probably want to use different terminology from 'pathogen' and 'host'
> when describing interactions (metagenotypes) that aren't pathogen-host interactions.

I don't think there is a link to get to the metagenotype details page for an interaction so this might not be an immediate problem.

jseager7:
@kimrutherford I've pushed some changes that move the metagenotype display name to a subheading and replace the page title with something shorter and more generic. Here's a screenshot from my local session:

image

The changes to the description section are now in their own issue (#2220), so this is ready to be merged.

kimrutherford:
That's great.  Thanks.

PR #2197, opened 2020-01-27
Rename 'Back' button to 'Cancel' on gene upload page

### LINKED ISSUE #2181: Rename the 'Back' button on gene entry page
Related: #1624 

When you first start a session and reach the gene entry page, the only navigation button available is a 'Continue' button, which is disabled until you've entered at least one gene ID or specified that the paper has no genes:

(Note the image shows PHI-Canto's version of the gene entry page, but I believe it works the same way for all versions.)

Then, when you return to the gene entry page after having entered genes, a second navigation button appears, labelled 'Back':

The Back button is always enabled, but the Continue button doesn't enable until the user has entered a gene (or picked a host organism in the case of PHI-Canto).

The problem is that the Back button does almost exactly the same thing as the Continue button: they both navigate to the same page; the only difference is that the Continue button stores your changes, whereas the Back button doesn't. If the Back button truly navigated the user backwards in the workflow, it would take them to the page where they entered their curator details. So, the Back button is redundant from a navigational perspective. I think the page would make more sense if the Back button was replaced with a Cancel button (as shown below), since this matches the intended behaviour of the button.

jseager7:
Fixes #2181 

This PR changes the label of a navigation button on the gene upload page from 'Back' to 'Cancel', since the navigation doesn't navigate backwards in the workflow, it simply cancels adding genes.

The buttons are generated by `Curs.pm` in a very dynamic and generalised way; so this wasn't a simple case of editing a template. My changes involved adding a special override to the general button generation code that changes the style class when the button name equals 'Cancel'.

I think I've replaced all references to 'Back' in `sub gene_upload`, but I'm not certain. I'd appreciate some double-checking to make sure I haven't messed up any logic by accident.

**Suggestion:** It might be more maintainable to simply replace the substring `"btn-primary"` with `"btn-warning"` rather than setting the whole string ([see here](https://github.com/pombase/canto/compare/master...jseager7:2181-gene-page-cancel-button?expand=1#diff-a15d4a21daa64924f075539239ed6014R775)), but I wasn't sure of the best way to do this; I figure Perl has a lot of syntaxes for replacing text, and I'm not sure which is preferred.

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -700,7 +700,7 @@ sub gene_upload : Chained('top') Args(0) Form
       map { [ $_, $_ ] } @{$c->config()->{curs_config}->{no_genes_reasons}} );
 
   if ($st->{gene_count} > 0) {
-    push @submit_buttons, "Back";
+    push @submit_buttons, "Cancel";
   } else {
     @no_genes_elements = (
       {
@@ -771,6 +771,9 @@ sub gene_upload : Chained('top') Args(0) Form
         if ($_ eq 'Continue') {
           $ret->{attributes}->{'ng-disabled'} = '!isValid()';
         }
+        if ($_ eq 'Cancel') {
+          $ret->{attributes}->{'class'} = 'btn btn-warning curs-finish-button button';
+        }
         $ret;
       } @submit_buttons),
         @no_genes_elements;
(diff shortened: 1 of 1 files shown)


kimrutherford:
Thanks for the PR it all looks good.

> It might be more maintainable to simply replace the substring "btn-primary" with "btn-warning"
> rather than setting the whole string 

This should do the trick:
``` 
$ret->{attributes}->{'class'} =~ s/btn-primary/btn-warning/g;
```

kimrutherford:
> This should do the trick:

Sorry, please ignore me.  I now see that you've done that already.

PR #2207, opened 2020-01-28
Add shortcut to restart session approval on front page

### LINKED ISSUE #2186: shortcut to restart approval
If you have a lot of edits to do in a list of sessiosn, it is quite cumbersome to access each session.

 From the PubMed search,
you need to go "view session"
"Finish reviewing"
"restart session approval"

Would it be possible to have an additional button   direct to "restart session approval" for admin?

jseager7:
Fixes #2186 

This pull request adds an extra button after searching for a publication in admin mode that provides a shortcut to restarting approval of a session:

image

The link only shows if the user is an admin, and if the publication has a session assigned to it.

I noticed that `CantoGlobals.current_user_is_admin` isn't defined on the front page, so I swapped it out for `CantoGlobals.is_admin_user`, which is defined. Not sure if this indicates a wider problem that needs to be addressed.

Also, the new code doesn't check that the chosen session is actually in approval mode before restarting it, but as far as I can tell this doesn't make a difference, because for sessions that aren't in the correct mode, the link doesn't have any (obvious) effect anyway.

If you think that the extra check for `$scope.userIsAdmin` in `$scope.restartApproval` is needlessly paranoid, then I'm happy to remove it. The button doesn't even get included in the DOM if `$scope.userIsAdmin` is false (due to `ng-if`).

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -1441,9 +1441,8 @@ var pubmedIdStart =
 
         $scope.restartApproval = function () {
           if ($scope.userIsAdmin) {
-            window.location.href =
-              CantoGlobals.application_root + '/curs/'
-              + $scope.data.results.sessions[0];
+            window.location.href = CantoGlobals.application_root + '/curs/'
+              + $scope.data.results.sessions[0]
               + '/restart_approval/';
           }
         };
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -1383,7 +1383,7 @@ var pubmedIdStart =
           searchId: null,
           results: null,
         };
-        $scope.userIsAdmin = CantoGlobals.current_user_is_admin;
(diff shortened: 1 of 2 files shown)


root/static/js/canto-modules.js
```diff
@@ -1383,7 +1383,7 @@ var pubmedIdStart =
           searchId: null,
           results: null,
         };
-        $scope.userIsAdmin = CantoGlobals.current_user_is_admin;
+        $scope.userIsAdmin = CantoGlobals.is_admin_user == "1";
```
kimrutherford: At some point we should change all of the CantoGlobals var to boolean.  :-)

kimrutherford:
Thanks James.  Looks good to me.

root/static/js/canto-modules.js
jseager7: There's an issue open for this at #2209. Seems `is_admin_user` is the only property that has this problem at the moment.

PR #2211, opened 2020-01-30
Prompt users to add at least one pathogen gene

### LINKED ISSUE #2174: Warn users when no pathogen genes have been added
Related: #2153, #2134, #2171

We should amend the warning box on the confirmation page to show a message about the lack of pathogen genes (when applicable). I think this message should have priority over warning the user about the lack of strains. Here's a mockup:

image

jseager7:
Fixes #2174

This PR extends commit https://github.com/pombase/canto/commit/eadc8141dad3696f75778dd7ad6cbd3eba19d2eb to also warn users to add a pathogen gene before they can start a curation session. The user isn't warned about missing strains until they've added a gene; that prevents a bug from before where users would be warned about missing strains despite having no way to add them.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1686,12 +1686,12 @@ a:not([href]) {
   font-weight: 700;
   margin-bottom: 10px;
 }
-.curs-strain-notice-container {
+.curs-warning-container {
   display: flex;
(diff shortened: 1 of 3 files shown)


kimrutherford:
That works for me.  Thanks James.

PR #2224, opened 2020-02-06
Display metagenotype details as individual rows

### LINKED ISSUE #2220: Improve display of metagenotype description
From #2193

On the Metagenotype details page, we could break the Description section up to describe the constituent parts of the metagenotype, rather than simply repeating the display name:

image

This has the advantage that we can link to the genotype pages for the constituent genotypes, and italicise the species names.

jseager7:
Fixes #2220 

This PR updates the 'Details' section on the Metagenotype Details page, so that it displays the properties of the metagenotype's constituent genotypes individually, rather than just displaying the display name for the metagenotype:

2220-before-after

The new code also links to the genotype page for the constituent genotypes, but only if the genotype is not the wild-type genotype (the code checks for this by testing if `$genotype->allele_string()` is empty, but it might be better to have a property on `Genotype.pm` that stores whether the genotype has alleles or not).

The template reuses the `$_unicode_allele_symbol` subroutine to encode allele symbols in the details section.

In pathogen-host mode, the rows are titled as 'Pathogen' or 'Host', but when not in pathogen-host mode, the template will display Organism A, Organism B (and so on) instead. This means that the change is able to support multi-organism mode when needed:

image

Note that if `strain_mode` is not true, then the strain rows are hidden.

- - -

@kimrutherford This hasn't been manually tested in single-organism or multi-organism modes, but single-organism mode surely isn't applicable, and multi-organism mode I think can be simulated just by overriding the relevant configuration values (which I did).

Also, let me know if you're not keen on dumping all this logic in a template, rather than keeping it in a module somewhere. I was hoping it would be possible to pull more of the data out of the `$metagenotype` itself rather than resorting to using the lookup services, but if you're okay with that, then I am too.

### DIFF
--- a/root/curs/metagenotype_page.mhtml
+++ b/root/curs/metagenotype_page.mhtml
@@ -69,8 +69,13 @@ Details
 % } else {
           <td class="title">Genotype A</td>
 % }
-          <td><% $_unicode_allele_symbol->($interactor_a->{'genotype'}) | n %></td>
+          <td>
+            <a href="<% $genotype_link_a %>">
+              <% $_unicode_allele_symbol->($interactor_a->{'genotype'}) | n %>
+            </a>
+          </td>
         </tr>
+
         <tr>
 % if ($pathogen_host_mode) {
           <td class="title">Host</td>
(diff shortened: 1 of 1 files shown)


jseager7:
Not sure why Travis-CI is failing, because it passed on my fork.

kimrutherford:
> Not sure why Travis-CI is failing, because it passed on my fork.

It's the same intermittent PubMed lookup failure.  I'll have to have a look at that.

Your PR looks great and works for me.

> Also, let me know if you're not keen on dumping all this logic in a template

I think it's fine as it is.

Cheers.

Issue #1945, opened 2019-06-12
small changes to genotype management pulldown menu


ValWood:
This menu isn't very popular, because if you don't have mouse control and go completely horizontally, the menu "slips" to the next gene.

We don't have any practical ~selections~ suggestions for an alternative.  However, we think a couple of small improvements might help a little.

 - [x] If at all possible can we increase the font size slightly here? Is the font size the same as the table, or slightly smaller? 

- [x]  Swap the order of the options so that the most frequently used are at the top.
This would be:
i) copy Edit
ii) edit details
the rest

kimrutherford:
> This menu isn't very popular

I don't like it because I think it's odd to have a menu that pops up when you mouse-over.

> We don't have any practical selections for an alternative.

We could go back to how it was originally.  We had an "actions..." link at the end of each row.  When you click the link, the menu pops up.

ValWood:
We talked about a similar option whereby if you click on a row the pop-up menu is stably associated with the row until used. This invokes an additional click and was vetoed by some because of this.  The old "actions button" also requires an additional click. But maybe this is better. We can discuss today.

kimrutherford:
I'd prefer a click either in the row or on a link to get the menu because it seems more normal to me.  A link (like "actions...") is more obvious as it makes it clear to the user that they can click.

jseager7:
My suggestion isn't really a small change, but have you ever thought about using icons for the row actions instead? Here's a mockup:

![image](https://user-images.githubusercontent.com/37659591/59425813-925f2880-8dce-11e9-93eb-5e8198e43a31.png)

This makes the actions less obtrusive. We can still add tooltips to explain what the buttons do, similar to what GitHub and other sites do for their icons:

![image](https://user-images.githubusercontent.com/37659591/59426090-2fba5c80-8dcf-11e9-819f-fd1b09909f4e.png)

We could also add text with the icon when the action needs to be more obvious:

![genotype_action_icons_text](https://user-images.githubusercontent.com/37659591/59426318-cc7cfa00-8dcf-11e9-959b-49bf353e78f7.PNG)

And because the icons can fit into the same vertical space as a row, we could have them visible all the time, meaning no extra clicks are needed. (Note the following example removes some styling to keep the interface less cluttered)

![action_icons_each_row](https://user-images.githubusercontent.com/37659591/59426892-3d70e180-8dd1-11e9-8b9e-54aa75adf969.PNG)

[...]

PR #2225, opened 2020-02-06
Small improvements to genotype row actions

jseager7:
Related: #1945 

This pull request changes the actions popup on genotype tables, specifically:

* increases the font size to match Canto's table data; and

* reorders the actions in the list so that the more frequently-used actions are closer to the top.

image

@ValWood I've followed your suggestion to place 'Copy and Edit' and 'Edit' as the second and third items in the list (respectively); for the rest of the items their ordering is unchanged, relatively speaking.

This change will affect all versions of Canto, so @mah11 @vmt25 are you happy with these changes, or would you prefer any other changes to the ordering of the actions list? I know there's a few actions not shown in the screenshot that are used by FlyBase (e.g. genotype notes).

(Note that PomBase isn't tracking the `master` branch of Canto, so even if these changes are merged, they might not appear on PomBase's Canto immediately.)

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -470,7 +470,6 @@ table.curs-genotype-list tbody {
 .curs-genotype-list-row-actions {
   border: 2px solid #bbb;
   padding: 3px;
-  font-size: 90%;
   z-index: 100;
   position: absolute;
   overflow: visible;
--- a/root/static/ng_templates/genotype_list_row_links.html
+++ b/root/static/ng_templates/genotype_list_row_links.html
@@ -1,8 +1,4 @@
 <div class="genotype-list-row-links">
-  <div ng-if="!read_only_curs">
-    <a ng-class="{disabled: genotypeId == null }"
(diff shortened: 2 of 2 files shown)


kimrutherford:
Hi James.  Sorry I haven't commented on this.  I think you should go ahead and merge the change.  We can tweak things later if needed.

kimrutherford:
> I think you should go ahead and merge the change.

Although I've only just noticed that there are conflicts.  Do you think that's easy to fix?

jseager7:
> Although I've only just noticed that there are conflicts. Do you think that's easy to fix?

I'll take a look. It's probably just because of extra items being added to the actions list in the meantime.

jseager7:
I've fixed the merge conflict. Here's how the list looks now, with the new 'View/edit notes' action enabled:

image

I also fixed a weird bug where the 'Add/edit background' link was being shown twice.

kimrutherford (approved):
Thanks James.  I think this can be merged.

PR #2228, opened 2020-02-17
Fix EditOrganismsSvc excluding hosts with no genes

### LINKED ISSUE #2172: Removing all host genes should remove host organism
In PHI-Canto, if you add a host organism to the session by adding a host gene, then delete that gene, the confirmation page makes it seem like the host organism has been deleted:

#### Before:

image

#### After:

image

But this isn't actually the case: the gene is deleted, but the host organism remains as a host with no genes. Once you return to the page, the host reappears:

image

This is inconsistent with how the pathogens behave: when you delete all their genes, the pathogen is deleted too, because PHI-Canto's curation rules prevent curation sessions without at least one pathogen gene.

I think we should either:

* **a)** remove the host organism when all of its genes are removed; or

* **b)** ensure that the page doesn't hide the host table when no genes exist, instead switching to showing a host with no genes added (as in the screenshot above).

- - -

Considering option B also made me realise that there's no prompt when deleting an organism (upon deleting the final gene). We might want to re-think this, because deleting an organism also deletes all its strains, including any custom strains that would have to be manually re-entered if an organism were deleted by accident.

So for this other problem, we could either: 

* **c)** warn the user that the organism and its strains will be deleted upon deleting the final gene, and prompt for confirmation; or 

* **d)** change the confirmation page so that deleting all the genes of an organism doesn't delete the organism (regardless of whether it's a pathogen or host), but leaves it as an organism with no genes, after which the delete button will remove the organism instead (though we should probably still prompt in this case for the sake of the strains).

I think option D is only worth doing if the user wants to 'amend' a gene by removing it and adding another one (presuming they added a gene in error). The user would still be able to do this under option C by adding the new gene first and removing the old one afterwards, but that might seem less intuitive.

- - -

See also: a54bd8c6462097bfa255763ec05104380ee95383 (from #1514), for a similar fix about deleting pathogen organisms.

jseager7:
Fixes #2172 

This PR makes a small change to `unsetOrganismGene` (in `EditOrganismsSvc`) so that it doesn't filter out organisms with no genes when said organism is a host.

@kimrutherford While working on this, I noticed something odd in `EditOrganismsSvc.removeGene`. Why is it necessary to remove a gene by doing a raw `$http.get` to another page in the application? Isn't there a service we can use for this?
```js
var url = curs_root_uri + '/edit_genes?gene-select=' + gene_id + '&submit=Remove selected';
$http.get(url).then(function () {...})
```

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -9278,16 +9278,14 @@ canto.service('EditOrganismsSvc', function (toaster, $http, CantoGlobals) {
   };
 
   vm.unsetOrganismGene = function (organisms, gene_id) {
-    organisms = organisms.map(function (o) {
+    return organisms.map(function (o) {
       o.genes = o.genes.filter(function (g) {
         return g.gene_id !== gene_id;
       });
       return o;
+    }).filter(function (o) {
+      return o.pathogen_or_host === 'host' || o.genes.length > 0;
     });
-    organisms = organisms.filter(function (o) {
-      return o.genes.length > 0;
-    });
-    return organisms;
   };
 
   vm.unsetHostOrganism = function (organisms, taxon_id) {


kimrutherford:
Thanks James.  The change looks good to me.

> Isn't there a service we can use for this?

There isn't a service currently.  It would be better to have a Service than submitting a form.  I've made a separate issue for that: #2229

Issue #2233, opened 2020-02-19
Encoding errors in PHI-Canto strain list


jseager7:
I've noticed that some of the strains that we've loaded into Canto haven't got the correct encoding.

Examples include the _Gallus gallus_ breed White Leghorn changing from 'weiße Leghorn' to 'weiÃe Leghorn', and the delta symbol getting mangled, such as 'G217B ura5Δ' changing to 'G217B ura5Î'.

I checked the database itself and it looks like these encoding errors are in there too:

![image](https://user-images.githubusercontent.com/37659591/74834336-22b51c00-5313-11ea-8ff5-2125eb4758ba.png)

Some of these examples will get cleaned up because the names will be localised to English, but that won't cover all cases.

@kimrutherford I'm keen to avoid this happening again when I update the strain list. Do you have any idea what might have caused the encoding problems, or how we can prevent it happening?

kimrutherford:
Could you attach the input file that is causing this or let me know where an example is?  I'll do some local testing.

jseager7:
The input files are on our (private) config repository. See here:

https://github.com/PHI-base/config/blob/master/data/

You'll want to download both strain files. You can see an example of a strain that has this encoding issue here:

https://github.com/PHI-base/config/blob/master/data/host_strains.csv#L677

kimrutherford:
I've just tried this locally and the characters made it through to the database.  Maybe this fix solved two problems?: https://github.com/pombase/canto/pull/2238#issuecomment-614035587

Could you test again when you get a chance?

jseager7:
I've tested the changes, and while 'weiße Leghorn' is a synonym now (so I can't see it in the interface), there are other examples of UTF-8 characters that are working correctly (see below).

![image](https://user-images.githubusercontent.com/37659591/79458782-3961b180-7fea-11ea-8fea-9baf0f677178.png)

I'll see if I can mark PR #2238 as fixing this issue.

jseager7:
> I'll see if I can mark PR 2238 as fixing this issue.

Turns out I can't do that without editing the original comment of the PR, so I'll just remember to close this after it's merged.

kimrutherford:
Can this be closed now?

Issue #2232, opened 2020-02-18
Add strain abbreviation column to Track schema


jseager7:
Related: #2110 

@kimrutherford In order to display abbreviated strain names, I think we'll need a new column in the `track` schema, in the `strain` table, that stores an abbreviated name for the strain. For example:

strain_id | organism_id | strain_name | abbreviation
-- | -- | -- | --
770 | 4565 | cv. Chinese Spring | CS

This abbreviation is supposed to be like a 'default' short name for display purposes in feature tables (as described in #2110). However, there's a problem that sometimes this short name may overlap with an existing synonym.

If we're not keen on enforcing defaults, another option is to shift this stuff to the `curs` schema, and let curators pick a display name from the list of synonyms instead, meaning abbreviations will be stored on a per-session basis. Then, we can define abbreviations as extra synonyms without needing to add another column.

This is lower priority than adding strain synonyms, so I think we should wait until we're sure about how abbreviations should be stored before committing to anything.

kimrutherford:
@jseager7 

When a curator adds a new strain to the session (ie. one we don't have already), will they need to be able to add an abbreviation and synonyms?

jseager7:
> When a curator adds a new strain to the session (ie. one we don't have already), will they need to be able to add an abbreviation and synonyms?

This might be too much work to force onto community curators – they probably won't have any need to specify synonyms because they'll surely choose their preferred name when adding a new strain. Admin curators should just edit the master strain list and add the abbreviations and synonyms there.

Allowing community curators to add abbreviations might be more useful, but only to prevent them adding the abbreviation as the preferred name rather than the expanded name, in cases where the name is too long (though we should really discourage this).

Overall, I don't think it's worth worrying about this at the moment. In the case of a truly novel strain there probably won't be any widely-used abbreviations or synonyms anyway.

jseager7:
It would probably be more useful to have an admin-only feature that allows admin curators to actually update the `track` strain table with new strains, synonyms, and abbreviations; but again, that's not essential.

kimrutherford:
After discussion we've decided that we don't need a strain abbreviation column.  It's not easy to pick an abbreviation that will be obvious to all users so it's best that we don't try.

Issue #2231, opened 2020-02-18
Add strain synonyms to Track schema


jseager7:
Related: #1718

Now that the PHI-base strain list has been checked and cleaned up, I think we should start thinking about how to add strain synonyms to Canto's `track` schema.

@kimrutherford I think we need a new table, `strainsynonym`, that serves a similar purpose to the `genesynonym` table: a one-to-many mapping between a strain name and its synonyms. For example:

strain_id | organism_id | strain_name
-- | -- | --
191 | 9606 | Caco-2BBE cell line

synonym_id | strain_id | strain_name
-- | -- | --
13 | 191 | C2BBE1
14 | 191 | C2BBe 1
15 | 191 | Caco-2/BBe 1
16 | 191 | Caco-2/BBe
17 | 191 | Caco-2BBe
18 | 191 | Caco-2 BBe
19 | 191 | Caco2-BBE
20 | 191 | Caco2BBe
21 | 191 | BBE

For now the table will probably be empty, but it will be populated once we import the updated strain list.

jseager7:
You might want to bundle this schema change with #2232, providing we can make a decision about that quickly enough.

jseager7:
Also, I first have to export our existing strains from PHI-Canto, so I can ensure that any existing strains (from the old list) get renamed to match the new strain list – that way, we won't have to delete the existing strains and re-add them. So, we can afford to defer the schema changes until that's done.

kimrutherford:
Hi James.

> I think we need a new table, strainsynonym

Yep, I think that makes sense.

> You might want to bundle this schema change with #2232

That makes sense too.

I'll have a go at this on a branch and make a PR.

I've also made an issue about the file format: #2235

Issue #2236, opened 2020-02-24
Modify Perl code to support new strain column and synonyms


kimrutherford:
See: #2232 and #2231

We'll need to change these classes: Canto::Track::StrainLookup and Canto::Curs::ServiceUtils

Possibly Canto::Curs::StrainManager too.

kimrutherford:
Those changes are done now so we can implement #2237 after a discussion.

PR #2238, opened 2020-02-24
Add a strain abbreviation column and synonyms

kimrutherford:
Hi James.  I've done the schema change but I haven't written the code to update an existing database.  I'll do that tomorrow.  I'll let you know when there's something to test.

Refs #2231
Refs #2232


kimrutherford:
> I've done the schema change but I haven't written the code to update an existing database. 

I've updated the upgrade script to handle this schema change.  There's not much to test yet because nothing makes use of the new column and table.

kimrutherford:
Hi James.

The loading code will now populate the strainsynonym table from the input file.  eg.

```csv
 NcbiTaxSpeciesId,CommonName,StrainCultivar
 3702,thale cress,Columbia Col-0,COL0,"col-0,columbia-0,colcol-0"
 9606,human,CFBE41o cells
```

If there are multiple strains that field will need to be quoted.

As we discussed, the strain details and its synonyms are completely reloaded each time.  So any existing abbreviation and synonyms will be removed and replaced with what is in the file.  If the abbreviation column is empty, the abbreviation will be removed from the strain.  If the synonyms column is empty all existing synonyms will be removed.

If a strain is completely absent from the input file the strain will be left untouched in the database.  So cuurently there is no way to delete a strain.  We can talk about that.

I've updated the wiki: https://github.com/pombase/canto/wiki/StrainDataFile

I think this code is now OK enough for you to test when you get a chance.  You'll need to run upgrade_db.pl

kimrutherford:
After discussion with James, I've removed the "strain_abbreviation" column.  We'll add it back in future if it's needed.

kimrutherford:
Hi James.  I think this is ready to merge.  Would you be able to test upgrading the schema?

jseager7:
Sorry about the delay on this. I've tested the schema upgrade itself and that worked with no problems on my local copy. I haven't tested it on the production database yet, though.

jseager7:
Spoke too soon, I get an error when navigating to the edit_genes page:

`
DBIx::Class::Storage::DBI::_prepare_sth(): DBI Exception: DBD::SQLite::db prepare_cached failed: no such column: strains.sraing [for Statement "SELECT strains.strain_id, strains.organism_id, strains.strain_name, strains.sraing FROM organismprop me JOIN cvterm type ON type.cvterm_id = me.type_id JOIN organism organism ON organism.organism_id = me.organism_id JOIN strain strains ON strains.organism_id = organism.organism_id WHERE ( ( type.name = ? AND value = ? ) )"] at lib/Canto/Track/StrainLookup.pm line 93
`

Surely `strains.sraing` is a typo?

lib/Canto/TrackDB/Strain.pm
```diff
@@ -58,8 +57,8 @@ __PACKAGE__->add_columns(
   { data_type => "integer", is_foreign_key => 1, is_nullable => 0 },
   "strain_name",
   { data_type => "text", is_nullable => 0 },
-  "strain_abbreviation",
-  { data_type => "text\n       sraing", is_nullable => 1 },
+  "sraing",
```
jseager7, line 60: I think this is where the "sraing" value first appears. I'm guessing this isn't intentional?

kimrutherford:
> Surely strains.sraing is a typo?

Whoops!  Thanks for spotting that.  I've pushed a fix.

kimrutherford:
I've just pushed a fix to allow non-ASCII characters (like delta symbols) in strain names.

jseager7:
Thanks. I'll test the loading again tomorrow.

jseager7:
I've tested the loading and it seems all the strains loaded successfully this time. Thanks!

I've also checked them in Canto, and while we don't have any way to display the strain synonyms yet, it seems everything else still works fine. Seems the encoding issues are fixed as well, based on the lowercase Chi (χ) in this example:

image

I had one problem related to an obsolete taxon ID while doing this, which I've opened a new issue about at #2264.

I don't think there's anything left to stop this from merging.

jseager7:
Fixes #2233

jseager7:
Oh, I didn't realise you'd already done the required service changes for strain synonyms in https://github.com/pombase/canto/issues/2236. Let me know if you want me to test those changes too before merging this.

kimrutherford:
> Let me know if you want me to test those changes too before merging this.

I think we can go ahead and merge then fix problems later.   Thanks for all the testing.

Issue #2244, opened 2020-03-10
Allow transfer of interaction annotation to multiple features


kimrutherford:
See: https://github.com/pombase/canto/issues/1830#issuecomment-590839270

kimrutherford:
After discussion we decided to add this transfer feature for phenotype_interaction first as it's the simple case.  I hope I'm remembering that correctly.

I've started work on this (e9ca0fde6, #2245).  So far it only allows transferring the annotation to genotype A of an interaction (it was the easy case).  I've just added that change to the flybase-test Canto.

@vmt25
@gm119

vmt25:
> phenotype_interaction first as it's the simple case. I hope I'm remembering that correctly.

Yes. This should be the easiest to implement because there are no rules on which genotypes can go with which.

vmt25:
> I've started work on this ([e9ca0fd](https://github.com/pombase/canto/commit/e9ca0fde61b9e657464610840a97209cf162285d), #2245). So far it only allows transferring the annotation to genotype A of an interaction (it was the easy case). I've just added that change to the flybase-test Canto.
> 

Works great!

vmt25:
From today's VC with Kim and Gillian:

A partial implementation that allows transferring to other Interactor A genotypes while keeping the interactor B is working fine on https://curation.pombase.org/flybase-test/  (as in #2245) and would already provide some added value, so it would be good if pushed into production in Fly-Canto after a couple of changes.

<img width="833" alt="Screen Shot 2020-05-05 at 15 30 48" src="https://user-images.githubusercontent.com/33721108/81077898-78966a80-8ee5-11ea-948c-3832d8ea2f32.png">

a) change the title of the menu page to a more generic 'Transfer genetic interaction annotation'

b) show the annotation to transfer (with the title 'interaction annotation to transfer')
<img width="639" alt="Screen Shot 2020-05-05 at 15 32 47" src="https://user-images.githubusercontent.com/33721108/81078163-c4491400-8ee5-11ea-90e0-21d9b8cbd905.png">

c) then list the possible interactor A genotypes where to transfer (as is, but eliminating the genotype of the original annotation)

vmt25:
(For future implementation)
A more common/useful case would be to transfer an interaction annotation, where interactor B ois the the genotype to be changed.

For this, the menu/pipeline will need to select which genotype to change (interactor A or B) and, depending on the selection, list the  possible genotypes (excluding the corresponding genotype on the original annotation)

[...]

PR #2245, opened 2020-03-10
Implement annotation transfer feature for interactions

kimrutherford:
See #2244

This isn't ready for merging yet.


kimrutherford:
I chatted to @vmt25 and @gm119 about this.  We decided to merge this change because it's useful even though it's not finished.

Currently only interactor A can be changed when transferring annotation.  I'll work on an option for interactor B.

The change is now in Fly-Canto.

PR #2279, opened 2020-04-29
Remove 'Unknown' expression and use 'Not assayed'

### LINKED ISSUE #2215: Enable 'Not assayed' expression level on wild-type alleles
Previously, issue #2046 added an 'Unknown' expression level type for use in pathogen-host mode. Later, in issue #2112, this 'Unknown' expression level was also enabled for wild-type alleles in single organism mode.

Now, according to Val, it seems it would have been better to use 'Not assayed' for issue #2112 instead.

'Not assayed' is currently disabled in Canto's code when the allele type is wild-type. The documentation doesn't exactly say why this is the case, but it does mention the fact it's disabled:

> Expression level: You will be prompted to define the expression level relative to wild-type […] If the product level was not measured […] choose "Not assayed" […] If a wild-type allele was investigated in the absence of any other changes, then an altered expression level must be set.

@ValWood can you please confirm the following:

* Do you want 'Not assayed' as an expression option for wild-type alleles?

* Do you want 'Not assayed' enabled on just pathogen-host mode, or single-organism mode, or both?

* If 'Not assayed' is enabled on wild-type alleles, do you want 'Unknown' to be disabled for wild-type alleles so you don't see both options at once?

jseager7:
Closes #2215 

This pull request removes the 'Unknown' expression level option from the list of expression levels, since PHI-base and PomBase seem to have come to a joint decision that this expression level is redundant. This has the effect of reverting commits 5a201fd7f769356816d52ef8fddba8d33725cb71 and a117bce4ee80ef911bf300a78ef6b8069c866c7a.

I think PHI-base decided to use 'Not assayed' in place of 'Unknown' for wild-type alleles, and this diverges from PomBase (who disallow 'Not assayed' for wild-type alleles), so I've updated the configuration to preserve this distinction.

@kimrutherford I've tested this in both single-organism and pathogen-host modes and it works fine. You might want to cherry pick these changes onto the `pombase-release` branch, since PomBase doesn't want 'Unknown' expression shown either.

### DIFF
--- a/root/static/ng_templates/allele_edit.html
+++ b/root/static/ng_templates/allele_edit.html
@@ -219,11 +219,6 @@
               <input ng-model="alleleData.expression" name="curs-allele-expression" id="allele-edit-not-assayed"
                      type="radio" value="Not assayed" />
             </div>
-            <div ng-show="pathogenHostMode || alleleData.type == 'wild type'">
-              <label for="allele-edit-expression-unknown">Unknown</label>
-              <input ng-model="alleleData.expression" name="curs-allele-expression" id="allele-edit-expression-unknown"
-                     type="radio" value="Unknown" />
-            </div>
           </td>
         </tr>
       </table>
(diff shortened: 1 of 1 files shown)


kimrutherford:
Thanks James.  Looks good to me.

kimrutherford:
> You might want to cherry pick these changes onto the pombase-release branch, since
> PomBase doesn't want 'Unknown' expression shown either.

Thanks.  I've done that.

Issue #1417, opened 2018-04-24
Write PHI-Canto specific documentation


kimrutherford:
We'll need documentation for pathogen-host mode.

(Replaces https://github.com/PHI-base/canto/issues/8 )

jseager7:
As discussed in #1664, we need some way to add the documentation files to Canto without checking them in. The current PomBase docs are checked in as a mix of MHTML and Markdown in the `/root/docs/` directory.

kimrutherford:
I'm hoping that the bulk of the PHI documentation can be added to the Canto docs directory.  That definitely makes sense for generic pathogen-host mode docs (rather than PHI-base specific stuff).  It also makes sense for PHIPO documentation to be added to the main repository in /root/docs/

We should mostly be able to put PHI-base specific tweaks in canto_deploy.yaml

kimrutherford:
> We should mostly be able to put PHI-base specific tweaks in canto_deploy.yaml

After discussion, that's probably not true.  We'll make a separate phi-canto-docs directory and use the existing docs directory as a starting point.

Job for Kim: allow configuration of the docs directory location.

kimrutherford:
> Job for Kim: allow configuration of the docs directory location.

I've done a first pass at that now.

 - I've added a new config setting `base_docs_path`
   - for PHI-Canto set it to "phi_base_docs" in `canto_deploy.yaml`:
                  `base_docs_path: phi_base_docs`
 - As a starting point, I've copied the existing documentation to `canto/root/phi_base_docs` where it can be edited as needed
 - I've added a directory for PHI-Canto specific images: `canto/root/static/images/phi_base_docs`
 - Some of the doc files in `phi_base_docs` are currently symbolic links / shortcuts to the corresponding PomBase doc file (eg. the GO annotation doc)
    - the phenotype doc file will need to be replaced with a PHI-Canto specific file

James, we should have a chat once you've had a chance to try this.  I'm sure some tweaks will be needed.

jseager7:
Thanks for this. I think our 'Getting started' page is pretty much finished with no remaining comments, so I can look at converting that to MHTML (all our documentation is currently stored as Google Docs, for ease of collaboration).

jseager7:
[...]

PR #2286, opened 2020-05-19
Add PHI-Canto documentation pages and images

jseager7:
Related: #1417 

This pull request adds documentation specific to PHI-base. The documents were previously stored as Google Docs; they've been downloaded as Markdown using an add-on and converted to MHTML with scripting. So far two pages are done. I'll update this PR as more pages are finished, unless we merge before then.

### Progress 

- [x] Getting started (index page)
- [x] Creating alleles and genotypes
- [ ] Curating phenotypes


jseager7:
@kimrutherford I made some changes to the `.screenshot` selector in `styles.css` to set a max-width on our images, since I decided not to scale down the images manually with thumbnails.

See here: https://github.com/pombase/canto/pull/2286/commits/2fffbd62d3be8812800aaad7d8beeac73d82a949#diff-eeabeec674069e72e8e638ad22871b98R1429-R1430

However, the changes to the selector will affect your docs for PomBase as well, so I think this rule should really be restricted to the PHI-base documentation, maybe by setting another class on the top-most `<div>` for our pages:

```html
<div class="docs phi-base-docs">
```
I'm open to any other suggestions. Now that I've found an image processing library for Python (which I'm using to build the docs), it wouldn't be much effort to generate scaled versions of the images.

kimrutherford:
> class="docs phi-base-docs"

That sounds like a good fix.  You can add the class to every doc page by editing
`root/phi_base_docs/autohandler`:

```diff
diff --git root/phi_base_docs/autohandler root/phi_base_docs/autohandler
index 8d67f58cc..51253848d 100644
--- root/phi_base_docs/autohandler
+++ root/phi_base_docs/autohandler
@@ -43,7 +43,7 @@ $hide_header => 0
   </div>
 % }
       <div class="clearall"></div>
-      <div class="docs">
+      <div class="docs phi-base-docs">
 % $m->call_next;
       </div>
     </div>
```

I checked the pombe docs with your changes and everything else looks fine.

jseager7:
Thanks for the suggestion about using the autohandler. I hadn't considered it because I got confused by the fact that `index.mhtml` already has a `<div class="docs">` near the start of the file. I guess this is an accident?

https://github.com/pombase/canto/blob/b1c7b9c4a947f359ad750a967fef888ec242e695/root/docs/index.mhtml#L1-L9

kimrutherford:
> I guess this is an accident?

Yep.  :-)

I had a look and it's currently
```html
<div class="docs">
   <div class="docs">
...
```
That's a bit unnecessary.

I think the class in index.mhtml can be safely removed.

jseager7:
@kimrutherford I've made the changes to restrict the screenshot styling to PHI-Canto. I've left the extra bottom margin on all the screenshots for now, since I think it benefits all the documentation. If everything looks okay in single organism mode, then I think we should just merge these two pages, and I'll add the phenotype documentation in a separate commit when that's done.

kimrutherford:
> If everything looks okay in single organism mode, then I think we should just merge this pull request, 

Yep!  It all looks good for pombe.

Issue #2290, opened 2020-06-01
Show species names in Transfer dialog in multi-organism modes


jseager7:
Sub-issue of #1830 

For PHI-Canto, in the Transfer annotation dialog, it would be useful to have the species names displayed next to the feature name for gene and genotype annotations, because sometimes we have the same gene or allele on different organisms, and there's no way to distinguish them in this case.

The example below shows TRI5 genes from _Fusarium sporotrichioides_ and _Trichoderma arundinaceum_, but there's no way to tell which is which. (The annotation is being transferred from TRI5 on _F. graminearum_.)

![image](https://user-images.githubusercontent.com/37659591/82307746-afc24c80-99b8-11ea-9b35-370768f545c2.png)

The proposed fix would have it look like this:

![image](https://user-images.githubusercontent.com/37659591/82308650-ff554800-99b9-11ea-9ec2-f4a3b1467bc2.png)

Note that the species name should also be displayed in the title of the modal.

jseager7:
## Species names in modal header

For the modal header, the following code needs amending to show the term, species name, and extension. We probably want to rename `termAndExtension` to something generic, like `termDetails`.

https://github.com/pombase/canto/blob/18efd948d9de55d1bdf0a10ea373d70ee47fae31/root/static/ng_templates/annotation_transfer.html#L3-L9

The string is generated here:

https://github.com/pombase/canto/blob/b1c7b9c4a947f359ad750a967fef888ec242e695/root/static/js/canto-modules.js#L7054-L7060

It looks like `annotationTransferDialogCtrl` already injects `CantoConfig` and `Curs`, so we have the means to check which mode Canto is in, and we also have another way to retrieve the organism details if they're not contained in the data for the feature.

https://github.com/pombase/canto/blob/b1c7b9c4a947f359ad750a967fef888ec242e695/root/static/js/canto-modules.js#L7039-L7043

## Species names in feature chooser

This case is more tricky, since the `multiFeatureChooser` directive that is used by `annotation_transfer.html` is very dynamic, and it always gets the feature name from the `feature.display_name` property:

https://github.com/pombase/canto/blob/18efd948d9de55d1bdf0a10ea373d70ee47fae31/root/static/ng_templates/multi_feature_chooser.html#L8

My ideas are as follows:

[...]

PR #2291, opened 2020-06-02
Show species name when transferring genotype annotations

jseager7:
References #2290

This pull request ensures that species names are shown in the Transfer Annotation dialog when transferring genotype annotations in multi-organism mode. Previously, the organism names were only shown for genes and metagenotypes.

This is how the transfer dialog looks now:

feature-chooser-transfer

This PR changes the interface to `featureChooserFilter` to accept a `showOrganism` argument, but it defaults back to the old behaviour when not set. The new behaviour is currently restricted to the `multiFeatureChooser` directive, so the organism name doesn't show in cases where it would be redundant – for example, in the Edit Annotation dialog:

feature-chooser-edit

@kimrutherford I haven't tested this in single organism mode, but I've checked all the possible combinations for the new logical expression and it seems sound (see below for a table of combinations that show the organism name). I haven't changed the condition that the organism name only shows in multi-organism mode.

- - -

multi_organism_mode | gene_id | genotype_id | showOrganism | Result
-- | -- | -- | -- | --
TRUE | **FALSE** | TRUE | TRUE | TRUE
TRUE | TRUE | **FALSE** | **FALSE** | TRUE
TRUE | TRUE | **FALSE** | TRUE | TRUE
TRUE | TRUE | TRUE | **FALSE** | TRUE
TRUE | TRUE | TRUE | TRUE | TRUE

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -383,9 +383,13 @@ canto.filter('encodeAlleleSymbols', symbolEncoder);
 canto.filter('encodeGeneSymbols', symbolEncoder);
 
 canto.filter('featureChooserFilter', ['CantoGlobals', function (CantoGlobals) {
-  return function (feature) {
+  return function (feature, showOrganism) {
     var ret = feature.display_name;
-    if (feature.gene_id && CantoGlobals.multi_organism_mode) {
+    var showOrganismName = (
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks James.  That's a nice clean solution.

Issue #2085, opened 2019-10-23
Link to metagenotype details from metagenotype annotation rows


ValWood:
We can do this in PomBase Canto. It's a nice way to check that the phenotypes to a genotype align across experiments in a long table.

It is difficult to implement in PHI-Canto because the multispecies-genotype spans two table columns.

Ideas?

kimrutherford:
That's supposed to be the case.  Could you paste in a screenshot and the session ID?

ValWood:
from th e summary page

<img width="685" alt="Screenshot 2019-10-23 at 09 44 22" src="https://user-images.githubusercontent.com/7359272/67374874-bb13dd80-f579-11e9-8a66-cf606f038c86.png">

we have never been able to do this.

ValWood:
I will have a radical suggestion. I was going to think about it for . a while.

In this table we could merge  pathogen and host genotype into "multispecies genotype"

It *should* be OK becasue  this is a display only view?

ValWood:
@jseager7  are we still going to 

Parastagonospora nodorum -> P. nodorum

jseager7:
> are we still going to Parastagonospora nodorum -> P. nodorum

That's the plan. I think the pull request that applies this change to the summary page is ready to go, I just got distracted and forgot to push it to GitHub. I'll do that shortly.

ValWood:
OK, no hurry. I thought  we decided yes, but sometimes these things are in my head...

jseager7:
> In this table we could merge pathogen and host genotype into "multispecies genotype"

Do you mean by merging the 'Pathogen genotype' and 'Host genotype' columns? That could make the resulting column contain a lot of information. However, it might give us the opportunity to save space in the case where the genotype names are short, and keep the two genotypes on their own lines:

![image](https://user-images.githubusercontent.com/37659591/68863122-a3e28e80-06e6-11ea-9368-1f916242ff0b.png)

Unfortunately trying to keep genotypes on individual lines leads to excessively wide tables in some real-world examples:

![image](https://user-images.githubusercontent.com/37659591/68863453-50247500-06e7-11ea-8fc6-297210b383cf.png)

And if we wrap the genotype names to stop this, we just end up trading height for width:

![image](https://user-images.githubusercontent.com/37659591/68863553-8235d700-06e7-11ea-9abf-89c50509cdbf.png)

Although it may be the case that the merged column is easier to read despite this.

ValWood:
[...]

PR #2298, opened 2020-06-08
Link to genotypes from metagenotype annotation rows

jseager7:
This PR links the pathogen and host genotype names in the metagenotype annotation table rows to the genotype details pages for their respective genotypes. Previously, the genotype names were just plain text, which wasn't consistent with the tables for single-species phenotype annotations.

The code doesn't create a link if the host is wild-type; the details page is irrelevant in this case, because the genotype should never have any annotations.

Here's how it looks:

image

This feature wasn't requested, but I think it's complementary to issue #2085.

- - -

@kimrutherford I was pretty rigorous in making sure that this doesn't affect any mode except pathogen-host mode, hence the code below:

```js
        $scope.hasWildTypeHost = (
          $scope.annotation.feature_type == 'metagenotype' &&
          CantoGlobals.pathogen_host_mode &&
          isWildTypeGenotype($scope.annotation.host_genotype)
        );
```

However, I'm concerned about the fact we're hard-coding the `host_genotype` and `pathogen_genotype` properties in the first place. Isn't this code used for FlyBase's metagenotypes as well? If I remember correctly, the agnostic display mode for other tables is 'Interactor A' and 'Interactor B'. For non-pathogen-host modes, do the `pathogen_genotype` and `host_genotype` properties implicitly map to the first and second genotype (respectively) in a metagenotype? If not, the whole directive probably needs to be refactored to not assume pathogen-host mode.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -205,9 +205,8 @@ function isSingleLocusDiploid(genotype) {
 }
 
 function isWildTypeGenotype(genotype) {
-  return genotype.genotype_identifier.indexOf('wild-type-genotype') !== -1
+  return genotype.alleles.length == 0;
 }
-
 function getGenotypeManagePath(organismMode) {
   var paths = {
     'unknown': 'genotype_manage',
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -204,6 +204,10 @@ function isSingleLocusDiploid(genotype) {
   return Object.keys(diploidNames).length === 1;
 }
 
+function isWildTypeGenotype(genotype) {
+  return genotype.genotype_identifier.indexOf('wild-type-genotype') !== -1
+}
+
 function getGenotypeManagePath(organismMode) {
   var paths = {
     'unknown': 'genotype_manage',
@@ -7654,6 +7658,7 @@ var annotationTableRow =
         $scope.featureType = null;
         $scope.interactionFeatureType = null;
         $scope.showInteractionTermColumns = false;
+        $scope.hasWildTypeHost = false;
 
         CursSessionDetails.get()
           .then(function (sessionDetails) {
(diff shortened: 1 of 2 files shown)


kimrutherford:
> I was pretty rigorous in making sure that this doesn't affect any mode except
> pathogen-host mode, hence the code below:

That looks good to me.

> However, I'm concerned about the fact we're hard-coding the host_genotype and
> pathogen_genotype properties in the first place. Isn't this code used for FlyBase's
> metagenotypes as well?

The metagenotype table in the session are used for both interactions and pathogen/host metagenotypes.  There's a bit of shared JS code but mostly they are handled separately.

> If I remember correctly, the agnostic display mode for other tables is 'Interactor A' and
> 'Interactor B'. For non-pathogen-host modes, do the pathogen_genotype and host_genotype
> properties implicitly map to the first and second genotype (respectively) in a metagenotype?

They do map like that on the server, but in the client code we can mostly treat that as an implementation detail.  The server will return host_genotype/pathogen_genotype only for pathogen/host metagenotypes.

The server code knows a metagenotype is pathogen/host if the annotation_type configuration has a category of "ontology": 
```yaml
  - name: disease_formation_phenotype
    category: ontology
    feature_type: 'metagenotype'
```

If the category is "interaction" it will return genotype_a and genotype_b when we query for metagenotypes.  eg.
```yaml
  - name: genetic_interaction
    category: interaction
    display_name: 'genetic interaction'
    feature_type: 'metagenotype'
```

root/static/js/canto-modules.js
```diff
@@ -204,6 +204,10 @@ function isSingleLocusDiploid(genotype) {
   return Object.keys(diploidNames).length === 1;
 }
 
+function isWildTypeGenotype(genotype) {
+  return genotype.genotype_identifier.indexOf('wild-type-genotype') !== -1
+}
+
```
kimrutherford, line 210: I think this might be a more robust:

```suggestion
function isWildTypeGenotype(genotype) {
  return genotype.alleles.length == 0;
}
```

jseager7: Thanks. I suspected there would be a better way to test for wild type genotypes. I've committed these changes.

kimrutherford:
I think this is ready to merge.  Is there anything else to do here James?

jseager7:
> Is there anything else to do here James?

I don't think so. Feel free to merge this.

PR #2300, opened 2020-06-11
Upgrade strain picker to show synonyms and allow autocomplete

### LINKED ISSUE #2237: Modify front-end code to display strain synonyms
See: #2232, #2231 and #2236

jseager7:
Closes #2237 

This PR changes the strain selector on the Gene and host organism list page so that it is able to display strain synonyms, and can support autocomplete (like the host organism picker on the gene entry page).

It would be ideal if only matching synonyms are shown in the case where the query matches a synonym and not a primary name.

## Tasks

The following tasks are required:

- [x] Enable autocomplete (replace native selector with Bootstrap version).
- [x] Add support for user-specified strains.
- [x] Display strain synonyms for each entry.
- [x] Autocomplete on both strain name and synonyms.
- [x] Add button as alternative control to add a strain.
- [x] Restyle strain tags to look less like buttons.
- [x] Add support for Unknown strains.

And these tasks are optional, but would be nice to have:

- [x] Visually indicate novel strains added by the user.
- [ ] Don't wait for the server before updating added strains.
- [ ] Limit height of suggestions pop-up and use internal scrolling.
- [ ] Match from the start of each word (instead of from any prefix).
- [ ] Only show synonyms which contain the search query.
- [ ] Sort primary name matches above synonym matches.
- [ ] Disable or hide strains in the typeahead that are already added.
- [ ] Disable delete strain button when strains can't be deleted.


jseager7:
It's now possible to enter custom strains using the new typeahead system, but I think the solution needs more thought.

Currently, the typeahead will select the first strain that contains your search text when you hit <kbd>Enter</kbd>:

image

So, if you're trying to enter a custom strain that's a prefix of a strain in the list, <kbd>Enter</kbd> won't do what you want: it just selects the strain from the list. For example, if you try to enter 'APEC DE2' when 'APEC DE205B' is in the list, you can't enter 'APEC DE2' without first hitting <kbd>Esc</kbd> (or breaking focus) to close the typeahead.

I considered changing the default behaviour so that the first matching entry _isn't_ selected, but this makes it far too easy to enter custom strains: unless you explicitly focus the entry in the list with your mouse or keyboard, the input will add any strain you type as a custom strain, even if it matches a strain in the list. Hitting <kbd>Enter</kbd> in the following example will add '042' as a custom strain, which seems counter-intuitive:

image

I can think of the following alternative solutions:

* **Option 1**: Check whether the custom strain matches an existing strain, and if it does, add the existing strain instead.

   * This is probably the simplest to implement, but it might seem counter-intuitive to the user if they expect their text to be auto-completed when they hit <kbd>Enter</kbd>: they may end up adding prefixes as a custom strain when they intended to add the first thing in the list.

   * However, this is how typeahead works on most search engines (where typeahead is only used when selected), so it might be _more_ intuitive to users than the sort of typeahead that programmers are used to.

* **Option 2**: Add a button to the side of the input that adds the text as a custom strain.

  * This makes the action more explicit, but will probably be more awkward for keyboard users as you have to tab over to the button. Unfortunately, the typeahead already captures <kbd>Tab</kbd> as a key to select a match in the list, so you have to close the typeahead before hitting <kbd>Tab</kbd>.

* **Option 3**: Use a custom template for the match list that has an action at the top that explicitly adds the entered text as a custom strain.

  * This has the same benefit of making the action explicit, and could be the most intuitive for users if search text is bound to the text of the action as you type it, for example:

     image

  * It would probably take some work to ensure that the first match is always focused, but that a keypress of the up arrow will then focus the action to add a custom strain. We don't want to focus the custom action by default. 

@kimrutherford What are your thoughts?

jseager7:
Here's a proof of concept for displaying strain synonyms:

image

I'm not too keen on the visual noise of the 'synonyms' label, particularly when it appears after almost every item, but I haven't thought of anything better yet. When omitted, it's not too clear where one strain name ends and the other begins:

image

Maybe this could be fixed with some extra borders on the list items, but that might need a custom template for the whole typeahead list.

kimrutherford:
> What are your thoughts?

I'd probably vote for option 2 because it's the most explicit - but option 3 is quite snazzy.  Let's talk about it on the Monday call.

> Here's a proof of concept for displaying strain synonyms:

Looks good!

> I'm not too keen on the visual noise of the 'synonyms' label, 

I see what you mean but I think that would be fine as an initial implementation.

jseager7:
The new strain picker can now suggest matches on the strain primary name and the strain synonyms. Any of these matches are highlighted in the same way. Here's an example of this:

image

This could only be done by applying a custom filter function (`strainFilter`) on the source array, since using a pattern object for the filter expression would only return `true` if the search text was a substring of the primary name _and_ the synonym (see [AngularJS API: filter](https://docs.angularjs.org/api/ng/filter/filter)), whereas we needed a match on the primary name _or_ the synonym.

Note that it's necessary to add a (very) short wait before the typeahead appears now, because otherwise the filter will trigger before the `ng-model` for the search text updates (meaning the filter uses the text you typed before, not what you just typed). Waiting for 2 ms seems to be enough to avoid this problem.

(See https://github.com/pombase/canto/pull/2300/commits/759cb375ae59d814a4c500eb3966cf1f139febc1#diff-ca8343304a606f3aab655041e1397602R28)

jseager7:
I've made a few more changes so that the strain picker behaves more sensibly:

- the `strainFilter` function now returns early if no text has been entered,
- there's a placeholder of 'Type a strain name' in each input,
- the input now resets to empty after a strain is selected or submitted, and
- the typeahead no longer focuses the first item by default, so it's now consistent with the experimental conditions typeahead.

jseager7:
In response to feedback from the last meeting, I've added a button to submit a user-specified strain when using the mouse:

image

This submits anything the user has typed into the field, and has the same effect as hitting the <kbd>Enter</kbd> key.

jseager7:
@kimrutherford There's something I'd like clarified about the `Curs` service: when calling `Curs.add('strain_by_name', ...)`, does the back-end take care of checking if the strain is in the list of existing strains before deciding whether the strain is a custom strain or not?

Basically, if I try to submit a strain that's already in the list (treating it like a custom strain), will the back-end reference the strain name from the existing list, rather than using the custom name, or are all the strains added as names without any links to the Track database?

kimrutherford:
Thanks James.  I've done a bit of testing and it's all working for me.

> In response to feedback from the last meeting, I've added a button to submit a user-specified strain when using the mouse:

Perhaps the button label should be "Add new strain" to be more explicit?

Also I wonder if we should make the added strains look a bit less like buttons?  Here's a quick example:

strain-add-buttons-2


```css
.curs-strain-tag {
    margin: 2px 2px 2px 0;
    display: inline-block;
    background-color: lightgrey;
    color: black;
    font-size: 105%;
    margin-right: 0.4em;
}
```

kimrutherford:
> Basically, if I try to submit a strain that's already in the list (treating it like a custom strain),
> will the back-end reference the strain name from the existing list, rather than using the
> custom name, or are all the strains added as names without any links to the Track database?

Yep, you can submit an existing or new strain with `Curs.add()` and it will do the right thing.  If it's an existing strain then the TrackDB strain is linked to the CursDB.  If it's a new strain, the name is added to the CursDB.

Here's the code:
https://github.com/pombase/canto/blob/ba7d5545aabeff4aca968eeb294a8bc2ec286968/lib/Canto/Curs/StrainManager.pm#L110

The strain lookup code also knows how to deal with new and existing strains, more or less transparently.

jseager7:
> Perhaps the button label should be "Add new strain" to be more explicit?

I can try this, although I'm not sure if the meaning is too different to 'Add strain'. From a domain perspective, it is technically possible to type a strain that's already in the list and then click 'Add strain' (instead of pressing <kbd>Enter</kbd> – which, strictly speaking, doesn't add a 'new' strain (in the sense of a custom strain that is only linked to the CursDB). From a user perspective, every strain is a 'new' strain, because Canto prevents you from adding multiple copies of the same strain. Maybe it's worth asking the other curators about this.

> Also I wonder if we should make the added strains look a bit less like buttons?

I agree that the strain tags need a rework. Currently it's accurate to style them as buttons because that's what they are – the strain delete action occurs if you click anywhere on the tags, not just the cross symbol – but they shouldn't behave like this in the first place. I wasn't sure whether to save this for a future issue to avoid changing too much at once. However, it makes sense to do this at the same time as adding a visual indicator for which strains are custom and which aren't, so I might as well work on this now. I'll start by styling them all more neutral.

> Yep, you can submit an existing or new strain with Curs.add() and it will do the right thing.

It's good to know I don't have to worry about preserving this distinction in the front-end code. Although, I'll probably have to keep track of the strain status in order to apply the contextual styling described above.

jseager7:
I've revised the style of the strain tags so they look less conspicuous, and I've given them clearer separation between the deletion button and the strain name. Here's how it looks:

image

The strain tags are now marked up as [Bootstrap button groups](https://getbootstrap.com/docs/3.4/components/#btn-groups), and the delete icon now uses a glyph from [Bootstrap's Glyphicons](https://getbootstrap.com/docs/3.4/components/#glyphicons), rather than the letter X.

Note that the deletion button behaves like a button, but the strain name doesn't, despite looking like one. Specifically, the strain name is simply a `<span>` styled as a Bootstrap button (`.btn.btn-default`), with an additional `.fake` style class applied to disable all interactive effects (i.e. pseudo-classes like `:hover` and `:focus`).

image

This preserves the appearance of the strain name and the delete action being part of one tag, but avoids the unexpected previous behaviour of the strain being deleted when the strain name was clicked.

Since this solution reuses Bootstrap code with a small amount of overriding, it should be simple to tweak the size of the buttons, or apply a different theme for the user-entered strain tags (such as the [primary button style](https://getbootstrap.com/docs/3.4/css/#buttons-options)).

@kimrutherford Are you happy with this solution?

kimrutherford:
That looks great.  I think it's a big improvement.

jseager7:
The strain picker will now apply different styling to user-specified strains (meaning those that aren't in TrackDB) – they are highlighted orange, specifically with the `.btn-warning` style class.

image

The solution was a bit convoluted, and involved adding `is_custom` properties to the strain objects in the `strainPicker` controller, after they're retrieved from `StrainsService`. This is a bit wasteful; it would be better to cache this information in `StrainsService`, but I decided to put fixing it on hold until I can look at refactoring the whole of `StrainsService` to be more like a model service.

I wonder whether it would be better to restrict this custom styling to Admin mode, or to only show when the session is in review, because I don't think that the distinction between custom strains and preloaded strains is particularly meaningful to community curators. It might just end up confusing them.

- - -

I've also ensured that the 'Add strain' button stays disabled until you can add a strain (until after you've typed some text in the input):

image

image

jseager7:
@kimrutherford I think all of the important tasks are done now, so assuming you don't have any other requests for changes or don't need to test, feel free to merge this. I'll move the other optional tasks into a new issue once this is merged.

jseager7:
@kimrutherford I was going to say this is fine to merge, but I just realised I forgot to re-implement the option for adding an Unknown strain. I'll just fix this quickly.

jseager7:
I've added support for Unknown strains again. These strains are now added using another button, next to the 'Add strain' button:

image

Clicking the button adds a strain called 'Unknown strain' (which is the same naming convention as before). The button is disabled after the unknown strain is added, even if a strain with the name 'Unknown strain' is typed into the input and submitted that way. The button is enabled again if the unknown strain is removed.

image


I decided not to style the unknown strain as a custom strain, since I think that the strain has a clear meaning. It wasn't treated as a custom strain in the old implementation either.

@kimrutherford I'm assuming that Unknown strains don't have special handling in the database?

kimrutherford:
Thanks for all that work James.  It's a big improvement.  I'm going to merge now so Val and Alayne can try it out.  We can tweak it later if needed.

Issue #2307, opened 2020-06-22
New extension range type needed: metagenotypes


kimrutherford:
From: https://github.com/pombase/canto/issues/2296#issuecomment-644139226

> We've decided we need to add a new annotation extension for this. The extension will relate
> a pathogen-host interaction phenotype annotation to the metagenotype representing its control
> case.
>
> @kimrutherford In order to add this, I'll need a new range ID on the extension config
> for metagenotypes, analogous to what we already for genes (GeneID).

See also: #2306

kimrutherford:
I've added a new range type: MetagenotypeID

I'll work on exporting these ranges as JSON next: #2306

Issue #2296, opened 2020-06-08
Add 'compared to control genotype' extension


ValWood:
We have decided that we need an additional extension option.

We really need in some cases  to have some way to link an annotation to the control in the gene for gene experiments.

We decided that it would be better to introduce a mechanism to do this more generally using an annotation extension.

 The extension option would be 
"compared to control genotype" and the options in the pull down would be all of the genotypes in the session.

We can chat about this on the next call.

ValWood:
Note that this will probably be a configuration option required by most  species that use Canto for curation.  The only reason we never need it for  fission yeast phenotype is that  we always use the same strain.

jseager7:
Is this needed for PHI-Canto only?

jseager7:
Sorry, ignore that – I didn't read your comments properly. What I should've asked instead is whether PHI-Canto prompted the need for this annotation extension.

ValWood:
>whether PHI-Canto prompted the need for this annotation extension.

yes, other users so far haven't required it. I'm not sure what Flybase do

jseager7:
Extensions are never part of Canto's core configuration, and we don't currently have any shared extension config except the ad-hoc sharing of GO extensions between PHI-base and PomBase.

So, presuming the original comment refers to phenotype annotations, we can enable this extension 
on PHI-Canto as soon as we have a way to link an annotation to a genotype (presuming that's not possible already; I know it's currently possible for genes). Any other users of Canto can opt-in to the extension when they're ready.

mah11:
Note: this is veering towards the notion of "dependent phenotypes" that [OMP](https://microbialphenotypes.org/) uses - see PMID:25433798: "Finally, OMP employs the qualifier “altered_relative_to” for dependent phenotype terms, which denote a difference in phenotype between organisms of two different genotypes or the same genotype assessed in two different environments."

jseager7:
> Note: this is veering towards the notion of "dependent phenotypes" that OMP uses

That would imply we'd need to be relating one phenotype annotation to another phenotype annotation, rather than to a genotype. @ValWood is there always an annotation capturing the wild-type control phenotype? If so, I think reusing what OMP has (or at least following their example) would make more sense.

mah11:
[...]

PR #2308, opened 2020-06-22
Allow metagenotypes as extension ranges

kimrutherford:
This is a first attempt at this and isn't quite ready to review.  I'm still testing.

Refs #2307
Refs #2296


kimrutherford:
Hi James.  I've done some more testing and I happy to merge this now.  The change should be invisible in the interface until an extension is configured with a metagenotype as the range.

Currently exporting produces incorrect output for these new extensions but the rest of the JSON file is fine.  I'll fix that problem next: #2308

I tested with this extension configuration:
```
PHIPO:0000001    is_a    compared_to_control_genotype    MetagenotypeID    compared to control genotype    0,1    user
```

Issue #2312, opened 2020-06-29
Allow wild-type pathogens in metagenotypes


jseager7:
Related: https://github.com/pombase/canto/issues/2296

In order to properly support the curation of scientific controls for pathogen-host interactions, we decided it was necessary to allow wild-type pathogen organisms in a metagenotype, in the same way that we currently allow wild-type host organisms.

I think this is a fairly simple change for the user interface, but I'm not sure if there might be constraints on the back-end that restrict support for this feature to host organisms. I'll test allowing wild-type pathogens on the front-end to see what happens.

PR #2313, opened 2020-06-29
Support wild-type pathogens in user interface

jseager7:
Closes https://github.com/pombase/canto/issues/2312

This pull request adds support for wild-type pathogen genotypes in metagenotypes, which is needed for the curation of scientific controls in pathogen-host interactions.

@kimrutherford I think I've made all the necessary changes to the front and back end to fix this, but I'd appreciate it if you could double-check the changes to `_metagenotype_store` (in `Curs.pm`).

- - -

Initially, the metagenotype API didn't support storing pathogen taxon IDs or pathogen strains. When trying to store both a wild-type pathogen and a wild-type host, you got a generic error message in the toaster.

I made a bunch of changes to `_metagenotype_store` to handle pathogens the same way as hosts when it came to creating metagenotypes (see 85f40ed93), and now it's possible to create a metagenotype where the pathogen and the host are both wild-type:

image

You can also create a metagenotype with a wild-type pathogen and a mutant host. I'm not sure if this is sensible, but it's easier to leave it enabled than it is to restrict this case.

image

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -1924,6 +1924,8 @@ sub _metagenotype_store
   my $body_data = _decode_json_content($c);
 
   my $pathogen_genotype_id = $body_data->{pathogen_genotype_id};
+  my $pathogen_taxonid = $body_data->{pathogen_taxon_id};
+  my $pathogen_strain_name = $body_data->{pathogen_strain_name};
   my $host_genotype_id = $body_data->{host_genotype_id};
   my $host_taxonid = $body_data->{host_taxon_id};
   my $host_strain_name = $body_data->{host_strain_name};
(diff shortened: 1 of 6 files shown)


jseager7:
Note to self: the pathogen genotype column in the annotation table isn't respecting the link disabling logic from https://github.com/pombase/canto/pull/2298. I'll bundle this fix up with this PR if it isn't merged before then.

image

jseager7:
> Note to self: the pathogen genotype column in the annotation table isn't respecting the link disabling logic

This is fixed by https://github.com/pombase/canto/pull/2313/commits/c854e25102a5a14ea80018a794c9e64dade3a196.

jseager7:
Just force-pushed in order to trigger the Travis-CI build again, since it failed for seemingly no reason.

kimrutherford:
> I think I've made all the necessary changes to the front and back end to fix this, but I'd appreciate it if you could
> double-check the changes to _metagenotype_store (in Curs.pm).

Looks good!  It interface works for me and the code changes all look good.  Thanks!

PR #2326, opened 2020-07-22
Link to metagenotype details from metagenotype annotation rows

jseager7:
(References #2085)

This pull request adds a 'View metagenotype' link to the `annotation_table_ontology_row.html` template when the `feature_type` is equal to `"metagenotype"`. It's needed because the feature names in the row only link to the constituent genotypes, not the metagenotype itself. Here's how it looks:

image

I also changed the navigation on the metagenotype details page so it's possible to get back to the summary page from there. Previously, it only linked back to metagenotype management:

image

## Remaining problems

I've set the pull request as a draft because of the (minor) unresolved problems below.

### FlyBase

@kimrutherford As far as I can tell, this change only affects ontology annotations on metagenotypes. I'm not sure if FlyBase use this type of annotation, or whether they only use metagenotypes in interactions. It might be a useful option for them too but I'm not sure.

### Circular links on metagenotype details

The metagenotype details page uses the same template as the summary page to list the annotations, meaning the 'View annotations' link is still visible on the metagenotype details page. It's completely redundant here, as it only links to the same page. I'm not sure of the best way to fix this. I can think of two options:

* **Option 1**: Check the URL of the page to see if the template is being rendered on the summary page, and only show the link if so. This is probably fine if our URL format is stable.

* **Option 2**: Add a binding to the table directive that controls whether to show metagenotype view links, or – as a more general solution – use a binding to specify the context in which the template is being rendered (e.g. 'summary' or 'details'). There might be an established pattern for this contextual display problem, but I'm not aware of one.

Note that this problem also applies to genotype details, where you can click the genotype link in the annotation rows to link yourself back to the same page, but I feel like it's less conspicuous in that case.

### DIFF
--- a/root/curs/front.mhtml
+++ b/root/curs/front.mhtml
@@ -106,7 +106,7 @@ $total_annotation_count
 % }
 </div>
 
-<annotation-table-list></annotation-table-list>
+<annotation-table-list show-metagenotype-link="true"></annotation-table-list>
 
 % if ($annotation_count > 0) {
 <div id="curs-annotation-download">
--- a/root/curs/metagenotype_page.mhtml
+++ b/root/curs/metagenotype_page.mhtml
@@ -120,10 +120,15 @@ Details
   </div>
 </div>
 
-<div class="clearall"/>
+<div class="clearall"></div>
 
-<button type="button" ng-click="backToMetagenotypes()"
-        class="btn btn-primary curs-back-button">&larr; Back to metagenotypes</button>
+<button type="button"
+        ng-click="toSummaryPage()"
+        class="btn btn-primary curs-back-button">&larr; Go to summary</button>
+
+<button type="button"
+        ng-click="toMetagenotypeManagement()"
+        class="btn btn-primary curs-back-button">Metagenotype management &rarr;</button>
 
 <annotation-table-list feature-type-filter="metagenotype" feature-id-filter="<% $metagenotype_id %>"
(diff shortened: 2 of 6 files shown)


kimrutherford:
> I'm not sure if FlyBase use this type of annotation, or whether they only use metagenotypes in interactions.

They're just using interactions.

> Add a binding to the table directive that controls whether to show metagenotype view links

I think this is the best solution.  If we need to change to having a "context" for the table we can do that later.

> but I feel like it's less conspicuous in that case.

I agree.

I tested your PR and it all works for me.

jseager7:
I've implemented Option 2. There's now an optional binding called `showMetagenotypeLink` on `annotationTableList` that flows down to `annotationTableRow`.

The binding value defaults to `false` (boolean type), but this has to be converted back to a string (via interpolation) before it reaches `annotationTableRow`, because `annotationTableRow` gets its data through `$attrs` instead of `$scope`, and `$attrs` doesn't seem to work as expected when objects are supplied (using one-way bindings). I've added the following check to protect us against the truthiness of `"false"`:

```js
$scope.isMetagenotypeLinkEnabled = function () {
  return $attrs.showMetagenotypeLink == 'true';
};
```

I'm guessing this conversion could be avoided by making `annotationTableRow` an element directive (rather than an attribute directive as it is now), and passing everything in via `$scope`.

Currently the only place where `showMetagenotypeLink` equals `true` is on the summary page (front.mhtml).

I've tested this with other annotation types and everything seems to work as expected, so I'll merge now.

PR #2331, opened 2020-07-29
Allow choosing session organisms in extensions

jseager7:
Related: #2255 

To create a complete GAF file for PHI-Canto, we need a way to capture all the organisms involved for the GO term 'interspecies interaction between organisms' (GO:0044419) and its children. The current plan is to capture the taxon IDs for the organisms using one or more annotation extensions.

Partial support for the range types 'TaxonID', 'PathogenTaxonID' and 'HostTaxonID' was added in commit 47a4a0e42a626d57cf897bcf327a496ef51dfbc3, where the new types are an alias of 'Number'. However, recent discussion indicates that it would be better to use a feature chooser that suggests organisms from the session based on their species names, rather than expecting the curator to enter the taxon ID manually.

@kimrutherford I've done as much as I can on the back and front end to add support for this. `ExtensionConf.pm` now handles the taxon ID types specially, and the `extensionRelationEdit` directive uses these range types to fetch the correct organisms. I'd appreciate some double-checking of the `ExtensionConf.pm` changes, since I also refactored the big tree of if-else blocks.

Unfortunately, I'm now stuck because the `featureChooser` directive doesn't work correctly with organisms. The property names for organisms aren't consistent with other features: for example, organisms use `full_name` instead of `display_name`, and they have no `feature_id`. This means they're incompatible with `featureChooser` and `featureChooserFilter`.

I'm not sure of the best way to fix this. I can think of the following options:

* **Option 1:** Rename the relevant object properties in a containing controller before passing them to `featureChooser`. This is the simplest option, but doesn't help us if we need to use the organism feature chooser in a different context.

* **Option 2:** Change the organism properties to be the same as any other feature (giving them a `display_name` and a `feature_id` and anything else they need). I expect this will be disruptive to a lot of existing code that relies on the current property names.

* **Option 3:** Keep all the old properties, but add the ones needed for `featureChooser`. This introduces redundancy, but would be helpful if we have to use the organism feature chooser in a different context.

* **Option 4:** Use the `organismSelector` instead of `featureChooser`. I'm not sure if this will work correctly, since `organismSelector` may not pass back the right data (taxon IDs), but it could allow us to avoid the problems with `featureChooser`.

Note that if we adapt `featureChooser` to support organisms, then the `organismSelector` directive will probably become redundant, and we should consider replacing it with `featureChooser`.

### DIFF
--- a/lib/Canto/Config/ExtensionConf.pm
+++ b/lib/Canto/Config/ExtensionConf.pm
@@ -112,48 +112,38 @@ sub parse {
       map {
         if (/:/) {
           push @new_ontology_range_scope, $_;
-        } else {
-          if (/^(Number|TaxonID|HostTaxonID|PathogenTaxonID)$/i) {
-            if (!grep { $_->{type} eq 'Number'} @new_range_bits) {
-              push @new_range_bits, {
-                type => 'Number',
-              };
-            }
-          } else {
-            if (/^text$/i) {
-              if (!grep { $_->{type} eq 'Text'} @new_range_bits) {
-                push @new_range_bits, {
-                  type => 'Text',
-                  input_type => lc $_,
-                };
-              }
-            } else {
-              if (/^(Gene|FeatureID|GeneID|ProteinID|TranscriptID|tRNAID|SP.*)$/i) {
-                # hack: treat everything else as a gene (and normalise the case)
-                if (!grep { $_->{type} eq 'Gene'} @new_range_bits) {
-                  push @new_range_bits, {
-                    type => 'Gene',
-                  }
-                }
-              } else {
-                if ($_ eq '%') {
-                  if (!grep { $_->{type} eq '%'} @new_range_bits) {
-                    push @new_range_bits, {
-                      type => '%',
-                    };
-                  }
-                } else {
-                  if (/^metagenotype/i) {
-                    push @new_range_bits, {
-                      type => 'Metagenotype',
-                    }
-                  } else {
-                    die "unsupported range part: $_\n";
-                  }
-                }
-              }
+        } elsif (/^(Number|TaxonID|HostTaxonID|PathogenTaxonID)$/i) {
+          if (!grep { $_->{type} eq 'Number'} @new_range_bits) {
+            push @new_range_bits, {
+              type => 'Number',
+            };
+          }
+        } elsif (/^text$/i) {
+          if (!grep { $_->{type} eq 'Text'} @new_range_bits) {
+            push @new_range_bits, {
+              type => 'Text',
+              input_type => lc $_,
+            };
+          }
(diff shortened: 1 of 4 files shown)


kimrutherford:
Hi James.  Thanks for digging into this.  Your changes in `ExtensionConf.pm` look good to me.

I think Option 4 using `organism-selector` is probably the best plan.  So I made a PR (https://github.com/jseager7/canto/pull/3) for your PR.  :-)

The `organism-selector` needed a slight tweek so that it would actually show the selector in the case where there was only one organism.  The action was quite confusing otherwise.  To see what I mean, remove `always-show-select="true"` from the organism-selector directive in `extension_relation_edit.html`.

I slightly simpiflied the controller for `extension-relation-edit` by re-using `$scope.organisms` for `$scope.pathogens` and `$scope.hosts` here: https://github.com/kimrutherford/canto/commit/c13e15e3047439454c32faad5425f9f1c5e3134d#diff-0713e61c0538c493314ecca53b336392L2807

I've done some testing and this seems to work OK.  Let me know what you think.

jseager7:
@kimrutherford Thanks for the help. Your changes work fine for me.

The only thing that I changed was the binding type for `alwaysShowSelect`. If you use string bindings (`@`) when passing around boolean values, the boolean is passed as a string ("true") instead of a boolean value (`true`). When you try to set the binding to `false`, you'll actually pass "false", and `"false" == true`. Using object-type bindings like the one-way binding (`<`) avoids this problem. I think string bindings are mainly intended for when you need interpolation on the binding value (although I haven't consistently followed this rule myself).

If you must treat booleans as strings (as in [#2326](https://github.com/pombase/canto/pull/2326#issuecomment-664351173)), then about the only reliable way I found was to treat everything except "true" as false. In this case, that would've looked like this:
```js
! $scope.alwaysShowSelect == "true"
```

kimrutherford:
> The only thing that I changed was the binding type for alwaysShowSelect.

Thanks for the fix.

jseager7:
@kimrutherford Sorry I didn't notice this before merging, but what's the purpose of `$scope.finishWithOrganism(organism)` in the `extensionRelationEdit` controller? It seems to be unimplemented. I only noticed because of a TypeError in the web console.
```js
$scope.organismSelected = function (organism) {
  $scope.extensionRelation.rangeValue = organism.taxonid;
  $scope.extensionRelation.rangeDisplayName = organism.scientific_name;
  $scope.finishWithOrganism(organism);
};
```

kimrutherford:
Good catch.  The was supposed to be:
```javascript
$scope.finishedCallback();
```

but now I've thought it over, it's not needed and the extra option on `organism-selector` (`always-show-select`) isn't needed either. 

I've just made a PR with a fix: pombase/canto#2332

PR #2332, opened 2020-07-30
Remove unneeded code and call to non-existing function

kimrutherford:
Refs #2331

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -2835,7 +2835,6 @@ var extensionRelationEdit =
         $scope.organismSelected = function (organism) {
           $scope.extensionRelation.rangeValue = organism.taxonid;
           $scope.extensionRelation.rangeDisplayName = organism.scientific_name;
-          $scope.finishWithOrganism(organism);
         };
 
         $scope.termFoundCallback = function (termId, termName, searchString) {
@@ -4559,7 +4558,6 @@ var organismSelector = function () {
       organismSelected: '&',
       organisms: '<',
       initialSelectionTaxonId: '@',
-      alwaysShowSelect: '<',
       label: '@'
     },
     restrict: 'E',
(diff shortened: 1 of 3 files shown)


jseager7:
This fixes this error for me, thanks. Although the changes to `alwaysShowSelect` makes the modal look a bit odd when there's only one organism in the session:

image

Thinking about this, in the case where there's only one organism in the session, we shouldn't be using the extension at all. The definition of 'interspecies interaction between organisms' (GO:0044419) is "any process in which an organism has an effect on an organism of a **different** species." So it never makes sense to use the same species for the extension.

Maybe we should just show a message in the extension editing modal when there's only one organism in the session of the respective type (pathogen or host)? Ideally we'd disable (or not show) the extension in the first place, but I don't suppose that's possible for Canto to do at the moment.

kimrutherford:
> Thinking about this, in the case where there's only one organism in the session, we shouldn't be using the extension at all.

> Ideally we'd disable (or not show) the extension in the first place, but I don't suppose that's possible for Canto to do at the moment.

It would be possible to fix that.  We'd need to pass more information into `extensionConfFilter()`.

> Maybe we should just show a message in the extension editing modal when there's only one organism in the session of the respective type (pathogen or host), telling the curator that there's no valid organisms to add.

That's an easy quick fix.

ValWood:
Note, this is probably already clear, but we only need the box if there is >1 species option in the  host group if annotating a pathogen gene, or in the pathogen gene if annotating the host

jseager7:
> Note, this is probably already clear, but we only need the box if there is >1 species option in the host group if annotating a pathogen gene, or in the pathogen gene if annotating the host

@ValWood does this mean the extension that gets shown depends on the role of the organism you're annotating?

To explain, currently we have two separate extensions for adding an extension with a pathogen or a host: **'with symbiont species'** and **'with host species'**. Both of these are shown for the GO term 'interspecies interaction between organisms' (GO:0044419). 

Would you only want to show 'with symbiont species' when the annotated organism is a host, and only show 'with host species' when the annotated organism is a pathogen?

ValWood:
>Would you only want to show 'with symbiont species' when the annotated organism is a host, and only show 'with host species' when the annotated organism is a pathogen?

yes, that's correct.

jseager7:
I'm going to merge this pull request since we need the fix, but I'll open new issues for the problems discussed here.

PR #2340, opened 2020-08-10
Update titles on table sort buttons to reflect sorting state

### LINKED ISSUE #2336: Add sorting buttons to column headers
(Sub-issue of https://github.com/pombase/canto/issues/2310)

Sorting annotation tables on the summary page is currently done by clicking the text in the column header. Unfortunately, It isn't very obvious how to sort the columns (or that sorting is even possible) until you mouse-over the column headers.

We could make it more obvious how to sort the tables by adding a visual cue. I'd suggest an icon, which is always visible, that indicates whether the columns are sortable. Clicking the icon would change the sorting mode of the column. The icons could change their appearance depending on the sort order.

Here's an example from Wikipedia:

image

jseager7:
(Closes #2336)

This pull request changes the titles on sort buttons so that they better reflect the actions from clicking the buttons.

In the default state, the sort button title is 'Sort ascending'. After being clicked, the title changes to 'Reset sort order'. I don't know what the default sorting criteria is, but if it's creation time, it could be better to be explicit about that and change the label to 'Sort by creation time' instead. Personally I don't have any preference.

I copied the logic used to determine the sort button title from the logic used to display the correct sorting icon, and it seems to work as expected based on my testing.

- - -

I've also bundled a fix for a typo in a function name: `setDefaulSort` &rarr; `setDefaultSort`.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7728,7 +7728,7 @@ var annotationTableCtrl =
 
         $scope.setSortBy = function(col) {
           if ($scope.sortColumn === col) {
-            $scope.setDefaulSort();
+            $scope.setDefaultSort();
           } else {
             $scope.prevSortColumn = $scope.sortColumn;
             $scope.sortColumn = col;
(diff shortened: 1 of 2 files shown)


kimrutherford:
> it could be better to be explicit about that and change the label to 'Sort by creation time' instead. 

Thanks for the fixes.  I'm not sure what's best either.  I'm going to merge the PR and we can edit the titles later if needed.

Cheers!

Issue #2320, opened 2020-07-20
Capture figure number as a separate data field


jseager7:
PHI-base have decided it would be useful to have figure numbers curated separately from the annotation comments. This will need a new field to be added to the annotation creation process, and possibly a new column to the annotation tables.

@ValWood @CuzickA We'll need to discuss which step of the annotation creation workflow should capture the figure number. Maybe on the same page as the comments?

Also, if anybody's not keen on having another column shown in the table, we could just include the figure information in the comments column (as we do now on an informal basis), but always have it shown at the start of the cell, possibly with unique styling. Here's an example:

![image](https://user-images.githubusercontent.com/37659591/87933912-06660880-ca86-11ea-9efd-69a528dfdc5b.png)

Note that a secondary advantage of including figure numbers is to enable sorting by figure, as discussed here: https://github.com/pombase/canto/issues/2310#issuecomment-660929450

kimrutherford:
> PHI-base have decided it would be useful to have figure numbers curated separately from the annotation comments.

I've added the Figure field.  James, could you have a quick test when you get a chance?  No database upgrade is needed.

kimrutherford:
I forgot to say, this isn't available for PomBase yet.  I'll sort that out tomorrow.

jseager7:
Thanks for the quick turnaround. I'll test this as soon as I can.

> No database upgrade is needed.

Can you remind me why this is? Is it because additional annotation data is stored as YAML in the database?

ValWood:
does that also include in the "duplicate and edit" window?

kimrutherford:
> Can you remind me why this is? Is it because additional annotation data is stored as YAML in the database?

Yep, there's a "data" column in the annotation table for storing miscellaneous proprieties.  It's YAML.

kimrutherford:
> does that also include in the "duplicate and edit" window?

Yep!

kimrutherford:
> I forgot to say, this isn't available for PomBase yet. I'll sort that out tomorrow.

I've add a Figure column to the main pombase server too now.  And also the test server in case you'd like to try it first.

jseager7:
@kimrutherford There doesn't seem to be any way to add the figure number when first creating the annotation in the 'standard' annotation workflow (e.g. the workflow that starts from the annotation tables and genotype management pages). The Figure field only seems to exist in the modal that's used for Quick Add, Edit, and Copy and edit.

kimrutherford:
[...]

PR #2347, opened 2020-09-09
Add help text and placeholders for Figure field

jseager7:
(References #2320)

This PR adds some explanatory text and a `placeholder` attribute for the 'Figure or Table' text input. I've made these changes both to the step-based ontology workflow and the modal-based workflow. See below:

### Step-based workflow

image

### Modal-based workflow

image

Note that the advice is specific to PHI-base, but this shouldn't matter if the Figure field is specific to PHI-Canto. I think the advice is fairly sensible for any curation to follow, anyway.

### DIFF
--- a/root/static/ng_templates/ontology_term_comment_transfer.html
+++ b/root/static/ng_templates/ontology_term_comment_transfer.html
@@ -14,7 +14,7 @@ You have made the following annotation:
 
   <div class="curs-workflow-fig-or-table" ng-show="showFigureField">
     <label for="figure-field">Figure or table:</label>
-    <input id="figure-field" type="text" ng-model="figOrTable" placeholder="Figure 1, Table 1"/>
+    <input id="figure-field" type="text" ng-model="figOrTable" placeholder="e.g. Figure 1, Table 1"/>
     <p>Prefix figure numbers with 'Figure' and table numbers with 'Table'. Prefix supplementary figure and table numbers with an 'S', for example: 'Figure S1'.</p>
   </div>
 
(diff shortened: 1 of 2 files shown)


root/static/ng_templates/ontology_term_comment_transfer.html
```diff
@@ -14,7 +14,8 @@
 
   <div class="curs-workflow-fig-or-table" ng-show="showFigureField">
     <label for="figure-field">Figure or table:</label>
-    <input id="figure-field" type="text" ng-model="figOrTable"/>
+    <input id="figure-field" type="text" ng-model="figOrTable" placeholder="Figure 1, Table 1"/>
```
kimrutherford, line 17: Is it worth tweaking that to:
```
eg. Figure 1, Table 1, ...
```
?

I don't have a strong opinion.

kimrutherford:
Looks good to me.  Thanks James.

root/static/ng_templates/ontology_term_comment_transfer.html
jseager7: I think that's a better choice. At first I thought it would make the figure field inconsistent with the placeholders in the other fields, as shown below:

image

But the other placeholders are more like labels for what goes in the field, rather than examples, so this isn't really a sensible objection.

I'm usually not keen on being explicit about the fact placeholders are examples, but I think in this case it's possible that curators could mistake the placeholder text for text that has actually been entered &ndash; particularly when using Edit or Copy and Edit &ndash; so clearly marking the text as an example will solve that. I'll make this change. Thanks for the suggestion.

jseager7:
I've made the requested changes. Merging now.

PR #2356, opened 2020-10-20
Remove unused JS libs, reimplement jTruncate()

kimrutherford:
I've made the changed but haven't tested yet.  I'll leave this as a draft until I've done that.


kimrutherford:
This PR removes the angular-touch, jquery.shiftcheckbox, jquery.jtruncate and jquery.validate libraries and also the various shims for "placeholder".

I've pasted the jquery.jtruncate code into canto.js and removed the functionality that we don't use which gets it down to 31 lines.  I hope we can move this code to AngularJS at some point so we remove this code.

Everything seems to work so I'm going to merge.  If you see any problems, let me know and I'll get on them straight away.

Issue #2325, opened 2020-07-22
Can't access organism page in review mode


jseager7:
I've noticed that in PHI-Canto, once a session has been submitted (which changes the session to review mode), it's no longer possible to access the page where you would normally edit genes, organisms, and strains. This means you have no way to review all the strains and wild-type hosts added to the session.

Specifically, it's the link highlighted below that gets hidden:

![image](https://user-images.githubusercontent.com/37659591/88175210-74910380-cc1d-11ea-88d0-0d754739dcdc.png)

Rather than hiding the link, I think it would be better to keep the page accessible, but disable all the edit actions (e.g. adding and deleting strains).

It seems this problem also applies to PomBase's version of Canto, where it's not possible to access the gene list page in review mode. However, it's less awkward to hide the page in PomBase's case, because PomBase doesn't use strains, meaning there's always a way to see all the information on the page in other places (specifically the detail pages for each gene).

- - -

@ValWood @CuzickA Have you ever had problems with the fact that the organism editing page is hidden in review mode? Maybe this is a non-issue since the link is still visible when approving the session – it's only hidden from non-admin curators that are visiting a submitted session. Also, any strains or wild-type hosts that have been used in genotypes or metagenotypes are still visible from the summary page, so the only thing that's hidden is the unused data.

ValWood:
I had not noticed so defer to @CuzickA

CuzickA:
Yes I agree with you @jseager7 
To view the information I reactivate the session and then resubmit once I've finished.

jseager7:
Okay, I'll try to fix this. It's possible that the reason the page is hidden is because it doesn't support read-only mode yet, so the fix might not be as simple as just keeping the link visible. I'll investigate further.

jseager7:
I just tried to access the Gene and Host Organism List page in read-only mode by appending 'ro' to the URL. Unfortunately, I got a 'page not found' error message:

> Page not found: curs/c7313b53a1533f0c/edit_genes/ro

@kimrutherford It looks like the subroutine in Curs.pm for edit_genes is different from other pages that do support read-only mode with 'ro', like genotype_manage:

```pl
sub edit_genes : Chained('top') Args(0) Form
```
```pl
sub genotype_manage : Chained('top')
```
Is the `Args(0)` method attribute causing the edit_genes page to ignore any extra arguments at the end of the URL?

kimrutherford:
Hi James.

[...]

PR #2364, opened 2020-11-10
Add a /view_genes_and_strains route and template

kimrutherford:
Refs pombase/canto#2325

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -51,6 +51,7 @@ use Carp qw(cluck);
 use JSON;
 use List::MoreUtils;
 use Try::Tiny;
+use Data::JavaScript::Anon;
 
 use Canto::Track;
 use Canto::Curs::Utils;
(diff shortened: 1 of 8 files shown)


jseager7:
Thanks for all your help. Seems like the read-only mode for the page has everything I need, so I'll look at fixing the user interface.

jseager7:
@kimrutherford I've made the changes to the user interface so that when in read-only mode, the editing controls are hidden, the link to the edit_genes page is disabled, and the summary page link goes to the read-only version of the summary page.

Note that I haven't made any changes to the AngularJS controllers to prevent edit actions when the session is read-only: I'm relying on `ng-if` directives in the template to prevent any of that code being triggered in the first place. Is that your preferred approach, or would you like guards in the controllers as well?

- - -
read_only_organism_page

jseager7:
While testing, I noticed that there's another page (or route), called confirm_genes, that uses the edit_organisms directives. The user is sent to this page after adding genes or organisms on the edit_genes page. So with the addition of view_genes_and_strains, we now seem to have three different versions of what is effectively the same page.

As briefly discussed in https://github.com/pombase/canto/issues/2325#issuecomment-724388795, it would be nice if we could decide on a more generic name for this page that isn't dependent on the state of the curation session.

jseager7 (approved):

jseager7:
@kimrutherford Is there anything left to do before merging this? All the interface changes to the Genes and Organisms page are done, and I've fixed the somewhat-related issue of an incorrect link to the confirm_genes URL in commit https://github.com/pombase/canto/commit/5e548854a40476e31f926a0f1efb907f723977b7.

kimrutherford:
Thanks James.  I did a bit more testing and found that the new route wasn't working correctly in non-PHI mode.  It was redirecting to the front page.  I've fixed that and now I think it's ready to go.

PR #2367, opened 2020-11-25
Add an extension config column to constrain by annotation type

kimrutherford:
This is on a branch for now to allow for testing.  It works for me.  :-)

I've added the new column to the right of the table and it's optional so no changes are need to existing configuration file.

As an example, here's how to restrict has_input for GO:0016485 to biological_process only:

```
GO:0016485    is_a   has_input    ProteinID    processes    0,1    user    biological_process
```

There can be multiple annotation type names separated by "|".  eg.

```
GO:0016485    is_a   has_input    ProteinID    processes    0,1    user    biological_process|molecular_function
```

### DIFF
--- a/lib/Canto/Config/ExtensionConf.pm
+++ b/lib/Canto/Config/ExtensionConf.pm
@@ -70,7 +70,7 @@ sub parse {
       next if $line =~ /^#/;
 
       my ($domain, $subset_rel, $allowed_relation, $range, $display_text, $help_text,
-          $cardinality, $role) =
+          $cardinality, $role, $annotation_type_name) =
             split (/\t/, $line);
 
       if ($domain =~ /^\s*domain/i) {
@@ -175,6 +175,7 @@ sub parse {
         help_text => $help_text,
         cardinality => \@cardinality,
         role => $role,
+        annotation_type_name => $annotation_type_name,
       );
 
       if ($domain =~ /(\S+)-(\S+)/) {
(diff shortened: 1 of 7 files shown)


jseager7 (approved):

jseager7:
Thanks for this. The design of the change looks fine and I have no problems with the code. I haven't tested the change locally yet (and I am planning to), but since the changes should have no effect when used with the current configuration it's probably okay to merge sooner if you want.

jseager7:
I've now tested this on my local copy with the ontologies and configuration values from the production server and everything seems to work fine. I'm happy for this to be merged whenever.

jseager7:
> I'm happy for this to be merged whenever.

Having said that, I'm presuming there are no problems if the configuration file is missing the annotation_type_name column? Presumably this was already the case for the PomBase GO extension configuration that I was loading; I didn't notice any problems due to that.

vmt25:
> This is on a branch for now to allow for testing. It works for me. :-)
> 
> I've added the new column to the right of the table and it's optional so no changes are need to existing configuration file.
> 
> As an example, here's how to restrict has_input for GO:0016485 to biological_process only:
> 
> ```
> GO:0016485    is_a   has_input    ProteinID    processes    0,1    user    biological_process
> ```
> 
> There can be multiple annotation type names separated by "|". eg.
> 
> ```
> GO:0016485    is_a   has_input    ProteinID    processes    0,1    user    biological_process|molecular_function
> ```

This configuration would work fine for us.

> Having said that, I'm presuming there are no problems if the configuration file is missing the annotation_type_name column? Presumably this was already the case for the PomBase GO extension configuration that I was loading; I didn't notice any problems due to that.

I have the same thought as @jseager7

kimrutherford:
> Having said that, I'm presuming there are no problems if the configuration file is missing the annotation_type_name column? 

Yep.  If the line has no value in that column, it's ignored.  So the existing configuration files should work as-is.

kimrutherford:
> I'm happy for this to be merged whenever.

Thanks for testing.

Issue #2369, opened 2020-11-25
'Add another...' link is misleading for multiple-cardinality extensions


jseager7:
(Raised by @CuzickA)

When adding annotation extensions with cardinality greater than one, and a biological feature as the extension range, it's easy to misinterpret the meaning of the 'Add another...' link. In the example below, the 'Add another...' link adds another gene to the _session_, but it looks like it's there to add the other protein or binding partner for the extension.

(Obviously if you try to use the link to add the other range value, it probably won't work, because the gene that you'll try to add should already be in the session, and therefore can't be added.)

![image](https://user-images.githubusercontent.com/37659591/100241666-da72fc00-2f2b-11eb-9733-e4d395bb90ff.png)

We could remove the confusion by making the link text more clear. For maximum clarity, I'd suggest **'Add new gene to this session'.**

- - -

Related to the above issue, I've always thought it would be better if the number of input fields in the extension editing pop-up matched the cardinality of the extension: so if you have cardinality 2, you get input fields for both the required values, and the two extensions are automatically created. Here's an example, with the link text change from above also applied:

![image](https://user-images.githubusercontent.com/37659591/100242903-3722e680-2f2d-11eb-81fb-cd50b31162ea.png)

ValWood:
I'd agree to both of those.

ValWood:
@mah11  do you envisage any problems with these 2 changes?

mah11:
Fine with me.

jseager7:
I'll go ahead with amending the link text since that's quick and easy.

Repeating the input fields to match the cardinality of the extension is going to be more difficult, and how sensible this solution is depends on how strict we want to be about required extensions (see #2365): for example, it's fine if we use it for extensions where the cardinality must be either 0 or 2 &ndash; since you never want the curator to add only one instance of the extension &ndash; but for cases where the cardinality can be 0 or 1 or 2, then repeating the input field in the pop-up might mislead the curator into thinking they have to add two instances of the extension.

We might be able to get around this usability issue with more prompting (and blocking the curator where multiple extensions are required), but maybe it would be better to limit the above solution to the case where the only valid cardinality values are 0 and some _n_ > 1 (with no intermediate values).

jseager7:
I've discussed this with @kimrutherford and we agreed that adding extra input fields probably isn't worth the effort until we can be sure which extensions are actually _required_. However, we did come up with some extra ideas for increasing the usability of the extension editing process:

### Required extensions

[...]

PR #2371, opened 2020-11-30
Be explicit about what the 'Add another [gene] ...' link does

jseager7:
(See #2369)

This pull request amends the link text in the extension editing modal so that it clearly indicates that the link adds another gene to the session, rather than adding another feature to the extension (in the case of extensions with a range of GeneID and a cardinality greater than one).

**Before**

extension-gene-link-before

**After**

extension-gene-link-after

@kimrutherford I'd like some feedback on the following points:

* There are two other templates that share the 'Add another ...' wording: `annotation_edit.html` and `annotation_evidence.html`. In both cases it looks like these links mean 'Add another gene to this session'. Do you want me to amend all of these links to use the same text as `extension_relation_edit.html`?

* I've changed the three consecutive full-stops (at the end of the link text) to an actual ellipsis character. I chose not to entity-encode this as `&hellip;` for the sake of keeping the template clean (since my editor font can display the ellipsis), and because we seem to be serving our templates as UTF-8, so there shouldn't be any encoding issues. Please let me know if you have any problems with this approach.

### DIFF
--- a/root/static/ng_templates/extension_relation_edit.html
+++ b/root/static/ng_templates/extension_relation_edit.html
@@ -6,7 +6,7 @@
                      chosen-feature-uniquename="extensionRelation.rangeValue"
                      chosen-feature-display-name="extensionRelation.rangeDisplayName"
                      chosen-feature-id="rangeGeneId"></feature-chooser>
-    <a ng-click="openSingleGeneAddDialog()">Add another ...</a>
+    <a ng-click="openSingleGeneAddDialog()">Add new gene to this session…</a>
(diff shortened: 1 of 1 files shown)


kimrutherford:
> Do you want me to amend all of these links to use the same text as extension_relation_edit.html?

Thanks for spotting those.  I went ahead and changed them to match your change.

> so there shouldn't be any encoding issues. Please let me know if you have any problems with this approach.

I agree.  It's 2020 after all.  :-)

PR #2373, opened 2020-12-07
Fix invisible autocomplete results

kimrutherford:
Increases the z-index of the results so that they are drawn above dialog boxes.


kimrutherford:
This is a problem when using the "Quick add" dialog and also a problem for adding extensions.

kimrutherford:
I think I've found a better solution so I'll close this.

PR #2374, opened 2020-12-07
Fix missing autocomplete results in dialogs

kimrutherford:
(Replaces #2373)

After a bunch of Google searching I found a better solution, documented here: https://api.jqueryui.com/autocomplete/#option-appendTo

I've added a "ui-front" class to the dialog HTML and that seems to do the trick.   For now I've just added it to the two dialog templates that I think need it, but there might be other places.

### DIFF
--- a/root/static/ng_templates/allele_edit.html
+++ b/root/static/ng_templates/allele_edit.html
@@ -1,5 +1,5 @@
 <div class="curs-allele-edit">
-  <form name="cursAlleleEdit" class="curs-allele-edit-dialog modal-content">
+  <form name="cursAlleleEdit" class="ui-front curs-allele-edit-dialog modal-content">
     <div class="modal-header">
       <h4 class="modal-title">
         <span ng-if="alleleData.display_name && !isCopied">
--- a/root/static/ng_templates/annotation_edit.html
+++ b/root/static/ng_templates/annotation_edit.html
@@ -1,4 +1,4 @@
-<div class="curs-annotation-edit" ng-if="annotationType">
+<div class="ui-front curs-annotation-edit" ng-if="annotationType">
   <div class="modal-header">
     <h4 ng-show="newlyAdded" class="modal-title">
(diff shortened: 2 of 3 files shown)


jseager7 (changes requested):
I've tested this and it seems to work fine. Both the experimental conditions picker and the term lookup render in front of the modal. 

During testing I found that the allele name autocomplete is also affected by this bug, but I'm not sure how to fix it. Putting the `ui-front` class on the element containing the `input` element didn't seem to fix anything. Here's what it should look like:

image

You can get to this modal through the 'Other genotype' button on the Genotype management page.

There might be other autocomplete fields out there, but I couldn't locate them in the interface. I think that the following text can be used to locate code using the autocomplete in `canto-modules.js`:

```js
.data("autocomplete")._renderItem = function (ul, item)
```

Based on testing it seems the search pattern can be shortened to just `"autocomplete"`.

Using this search I found references to an autocomplete in the following directives:

- `alleleNameComplete`
- `featureComplete`
- `termNameComplete`

`termNameComplete` seems to be fixed already, and `alleleNameComplete` is described above, but I don't know where (or when) `featureComplete` appears in the interface.

kimrutherford:
> During testing I found that the allele name autocomplete is also affected by this bug,

Thanks for testing.  I've pushed a fix for that.

> Putting the ui-front class on the element containing the input element didn't seem to fix anything.

I added ui-front to the element with the modal-content class in allele_edit.html and that seemed to work.

> but I don't know where (or when) featureComplete appears in the interface.

I think it's only used in one place at the moment.  If you're editing an genotype-phenotype annotation there's a search button beside the genotype selector, which turns the selector into an autocomplete box.  I think this case is covered by the previous fixes.

genotype-search-1

jseager7 (approved):

jseager7:
Thanks for the fix. I've tested all of the remaining autocomplete fields and they all draw in front of the modal, as expected. I'll go ahead and merge this.

PR #2380, opened 2020-12-15
Fixes for jQuery and jQuery-UI code on admin pages

kimrutherford:
I've had a thrash around and managed to fix some of the problems in the admin interface - dialog weren't showing and autocomplete wasn't working.  I don't know what I'm doing so let's chat about this on Skype.

I added https://www.npmjs.com/package/webpack-jquery-ui to get things to work but it's probably not required.


kimrutherford:
I changed `require()` to `import` and it all seems to work just as well.  I've also rebased the branch as there were a few commits on the webpack branch.

jseager7:
Thanks for the help. I might do some testing to see how `webpack-jquery-ui` works on `canto-modules.js` before merging, for example replacing this:
```js
import 'jquery-ui/ui/widgets/autocomplete';
```
with this:
```js
import 'webpack-jquery-ui/autocomplete';
```

jseager7:
@kimrutherford I've tested using `webpack-jquery-ui` in place of the `jquery-ui` package and I haven't noticed any problems with it. I tested the ontology term autocomplete for annotations and annotation extensions, the experimental conditions autocomplete, and the autocomplete used in the genotype search. I've pushed a commit to make the suggested change.

jseager7 (approved):

jseager7:
Just had to rebase this branch onto `webpack` again to fix the fact that a previous merge removed tag-it from package.json. Should be fine to merge now.

jseager7:
@kimrutherford I was trying to see if we could get rid of the jQuery-UI CSS files in `head.html`, because the `webpack-jquery-ui` package requires the CSS from node_modules instead, meaning it's already bundled using webpack's `css-loader` and `style-loader` and doesn't need to loaded twice. Unfortunately, the interface doesn't display correctly when the styles are bundled. 

image

It makes no sense to bundle the CSS files while at the same time loading CSS files for a different version of jQuery-UI from a CDN. At the moment I'm not exactly sure what's causing the bug when the styles are bundled. Short of figuring out the root of the problem, I can think of a few quick solutions, none of them particularly good:

* Disable `css-loader` and `style-loader` so that webpack will (presumably) ignore bundling the CSS;

* Import jQuery-UI components directly, without `webpack-jquery-ui` as an intermediary, allowing us to exclude the CSS files from the import;

* Try using `file-loader` to emit the CSS as a file instead of embedding it as `style` elements, just in case something about the embedding is causing issues.

My suspicion is the style rules (or selectors) are probably different between jQuery-UI version 1.9.1 (CDN) and 1.12.1 (npm). I'll see what I can find out by inspecting the autocomplete.

jseager7:
I've compared the style rules when loading the CDN styles vs. the bundled styles, and it looks like the classes `ui-widget` and `ui-widget-content` are the problem: these are missing in the bundled styles. The styles exist in `theme.css` in the `jquery-ui` package, and `webpack-jquery-ui` doesn't include this unless you import `webpack-jquery-ui/css.js` (all that does is import `theme.css` from `jquery-ui`). I guess it's not included by default because they assume you'll be writing your own styles.

So the fix _should_ be as simple as adding one import:

```js
import 'webpack-jquery-ui/css';
```
Unfortunately this throws a bunch of errors because webpack is now trying to load PNG icons from jQuery-UI without an appropriate loader. Guess I'll be fixing that next.

It looks like the icons from the old jQuery-UI version are checked in at `root/static/css/images`; I'll try to configure `file-loader` to put the processed assets there, just as a proof of concept, but we might want to reconsider versioning the icons if they're part of the build process now – much like our bundles, this depends on the deployment method we choose.

jseager7:
After configuring `file-loader` I managed to get the CSS loading from the npm package. Unfortunately, the styles still don't look the same, but they're closer:

image

The icons still don't load either, which is probably due to the CSS using relative URLs, meaning Canto looks for the images at this address:

`http://<domain>/curs/images/ui-icons_444444_256x240.png`

Not sure if this has to be fixed with routing or whether it can be resolved from the repository file structure. If we don't actually need the icons, we might be able to tell webpack to ignore them using `null-loader` (deprecated in webpack v5, we're still on v4) or something similar.

kimrutherford:
> Unfortunately, the styles still don't look the same, but they're closer

It looks OK when I try it locally.  It doesn't look like your screenshot.

I couldn't see any missing icons either.  Could you give me an example?

jseager7:
> It looks OK when I try it locally. It doesn't look like your screenshot.

I haven't actually pushed any of my changes to this branch yet, which might be why you're getting different results. If you want the changes pushed so that we can both investigate, then I'm happy to do that.

kimrutherford:
> I haven't actually pushed any of my changes to this branch yet

Ah!  :-)

Thanks for keeping me up to date.

I'm happy to (try to) help if you can't get any further.  But there's no need to push to the branch if you're still working on it.

jseager7:
@kimrutherford I've looked at this again and I can't find any way to fix the previously mentioned CSS problems. I think the quickest fix is to not use `webpack-jquery-ui` (since we have no control over the CSS loading with that package) and instead import _only_ the .js files directly from the `jquery-ui` package – the legacy CSS files can still be loaded from a CDN, as we were doing before. So, this seems to be the only change that's needed:

```js
// canto.js
import 'jquery-ui/ui/widgets/autocomplete';
import 'jquery-ui/ui/widgets/tooltip';
import 'jquery-ui/ui/widgets/dialog';

// canto-modules.js
import 'jquery-ui/ui/widgets/autocomplete';
```

(Note that I can't find any examples of `jquery-ui/ui/widgets/tooltip` being used in the admin interface or elsewhere, so if you know where to look for this, please let me know and I'll test it.)

I've tested this locally, and it seems to work correctly, plus it has a look-and-feel that's close to what we have on the `master` branch. Unfortunately, there are a few minor graphical differences (and graphical bugs), probably owing to a mismatch between the jQuery UI version from the CDN and the script files I'm loading from npm. I've illustrated these below:

### Autocomplete

#### jquery-ui 1.9.2

image

#### jquery-ui 1.12.1

image

With the new code, the autocomplete is using a different background image, and the classes that control the border rounding are absent, so the borders are all square. I don't have any clear idea why the background image has changed, but I suspect it's due to different style classes being used in the source for v1.12.1.

### Dialog

#### jquery-ui 1.9.2

image

#### jquery-ui 1.12.1

image

With the new code, the close button for the dialog shows the inner text, the button has different borders, and the button icon is aligned wrong. I might be able to fix this with some CSS hacking, but overriding the existing rules could be tricky. Here's a proof of concept:

image

- - -

Given that loading `jquery-ui` seems to be more flexible than `webpack-jquery-ui`, I think I'd prefer to close this pull request in favour of applying the above changes to the `webpack` branch.

Ultimately, I'd much prefer to replace the jQuery UI widgets with Bootstrap UI components, since the functionality is almost identical, and I've had much less trouble with Bootstrap UI in the past. I just need to find time to try to re-implement the autocomplete. I'm not sure how feasible it is to use AngularJS on the admin interface, but I'm happy to work on implementing that.

jseager7:
@kimrutherford I think I completely broke this pull request by rebasing the `webpack` branch, but I think the changes here should be superseded by commit https://github.com/pombase/canto/commit/d003fa31ebe258e8d3401b047ac556ad0105ff29 on the `webpack` branch. The new changes are basically the same as the fixes I mentioned in my previous comment:

```diff
- import 'jquery-ui';
+ import 'jquery-ui/ui/widgets/autocomplete';
+ import 'jquery-ui/ui/widgets/tooltip';
+ import 'jquery-ui/ui/widgets/dialog';
```

I'll close this PR in favour of the other branch, but thanks for your help tracking down the problems.

PR #2381, opened 2020-12-15
Fix centring of loading spinner

kimrutherford:
I noticed that on the webpack branch, the loading spinner was positioned near the top left of the screen.  This seems to fix it.


kimrutherford:
I tested the PubMed ID lookup too and it seems OK with this fix.

PR #2385, opened 2021-01-05
Indicate number of required or available annotation extensions

jseager7:
(References https://github.com/pombase/canto/issues/2369)

This pull request adds more information to the extension picker list (extension_or_group_builder.html), such that the list now indicates, for each extension, how many instances of that extension are available or required.

The messages are initially hidden, except where the extension is required (that is, where the first cardinality in the range is greater than zero). This is because many extensions have cardinality ranges of [0,1] and it just clutters up the interface having messages of '1 more available' next to every extension.

For example, the following annotation extension has a range of [0,2], so the message is hidden:

image

If the range were to omit zero and just be [2], the code would treat the extension as required, and the message would change accordingly:

image

After one instance of the extension is added, the message changes to state that more instances are required (this would also happen if the range were [0,2] since 1 would not be valid in that case):

image

When there are multiple valid cardinality intervals, for example [0,1,2], the code uses a different message to indicate this: '_n_ more available'. In the example below, one instance of the extension has been added, which is in the valid range, and the curator is informed that one more instance can be added before the maximum cardinality (or more generally, the next interval) is reached:

image

## Implementation details

I added two cardinality status codes to enable this new functionality:

* **OPTIONAL**: Used when the cardinality range has a minimum of zero and no instances of the extension have yet been added. Also used whenever an asterisk is encountered in the cardinality range.

* **MORE_AVAILABLE**: Used when there are multiple consecutive intervals in the cardinality range (e.g. [0,1,2]) to prompt the curator that more extensions can be added, but they are not required.

The MORE_POSSIBLE status code has been removed, since its purpose is now handled by a combination of the two new status codes.

As part of the changes, it's now possible to define a range of '_n_ or more' with the syntax `n,*`, where _n_ is any number. I think the code even supports ranges of '_n_ or _m_ or more' with the syntax `n,m,*`, up to an arbitrary number of intervals.

Zero is also a valid range, both for the sake of completeness and because it seemed safer to handle zero explicitly. Zero cardinality [0] has the effect of disabling the extension, so it's not useful in practice.

Here's a table summarising which messages are shown (if any) under a number of possible extension ranges:

Cardinality range | Extension count | Message
-: | -: | --
`*` | _n_ | [none]
`0,1` | 0 | [none]
`0,1,2` | 0 | [none]
`0,1,2` | 1 | 1 more available
`1` | 0 | 1 required
`2` | 1 | 1 more required
`0,1,3` | 2 | 1 more required
`2,*` | 0 | 2 required
`2,*` | _n_ > 2 | [none]

The list of cases handled by the code isn't exhaustive, so its behaviour isn't defined in the case of nonsensical cardinality ranges like `0,0,0` or `7,3,1` or even redundant ranges like `0,*`. I think this would be better handled by validating the extension configuration on the server side.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1569,6 +1569,14 @@ a:not([href]) {
   color: grey;
 }
 
+.curs-cardinality-required {
+  color: #8C0000;
+}
+
+.curs-cardinality-available {
+  color: #4F4F4F;
+}
+
 .curs-extension-name-complete-hint {
   white-space: nowrap;
   padding-left: 1em;
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -2550,35 +2550,41 @@ var extensionOrGroupBuilder =
         };
 
         $scope.cardinalityStatus = function (extensionRelConf) {
-          var count = $scope.getCardinalityCount(extensionRelConf);
+          var extensionCount = $scope.getCardinalityCount(extensionRelConf);
           var cardinalityConf = extensionRelConf.cardinality;
-
-          if (cardinalityConf.length == 1) {
-            if (cardinalityConf[0] == '*') {
-              return 'MORE_POSSIBLE';
-            }
-
-            if (cardinalityConf[0] == count) {
+          var minCardinality = cardinalityConf[0];
+          if (minCardinality == 0) {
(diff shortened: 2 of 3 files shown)


kimrutherford:
I've done some testing and it all works very well.  Thanks!  I appreciate you digging into this issue.

Issue #2163, opened 2019-12-12
Can't choose condition suggestions with keyboard


jseager7:
This is a regression caused by #2158: now if you focus a condition in the autocomplete suggestion list and press <kbd>Tab</kbd> or <kbd>Enter</kbd>, the autocomplete will complete whatever you typed in the box, rather than completing the focused condition:

![image](https://user-images.githubusercontent.com/37659591/70712984-fced1480-1cdc-11ea-909a-63c82038b29d.png)

You can still pick suggested conditions using the mouse, but I'm marking this as high priority in case anybody else uses the keyboard for completing experimental conditions. If this proves to be a big problem, then we should revert #2158 for now.

jseager7:
I've pushed a quick fix for this. It's a pretty awful hack involving jQuery selectors, but I couldn't figure out anything better. Essentially we scan the suggestions list to see if the user has focused any terms before adding a term, and if there is a focused term, that gets added instead of what the user typed:

```js
if (!(that.options.autocomplete.autoFocus && that.tagInput.data('autocomplete-open'))) {
    var focusedItem = that.tagInput.autocomplete('widget').find('.ui-menu-item > .ui-state-focus');
    if (focusedItem.length > 0) {
        that.createTag(focusedItem[0].innerText);
    } else {
        that.createTag(that._cleanedInput());
    }
    that.tagInput.autocomplete('close');
}
```

Based on manual testing, it doesn't seem to break adding custom terms, nor does it break selecting terms using the mouse, so I'll close this issue as fixed.

@kimrutherford Next time you're around, it might be worth seeing if we can come up with a cleaner fix. There's probably a better way; it just depends if the effort is worth it.

kimrutherford:
> It's a pretty awful hack

I'm a fan of awful hacks that fix bugs.

> Next time you're around, it might be worth seeing if we can come up with a cleaner fix.

Sounds good.  Is this hack needed in the latest tagit.js?

jseager7:
> Is this hack needed in the latest tagit.js?

Yes, because it's a regression bug introduced by the fix in #2158, which is merged in the upstream tag-it.js repository (so the bug presumably exists upstream too). I couldn't see any fixes suggested for this on their tracker, so I based this fix on the one in #2162.

PR #2386, opened 2021-01-06
Load tag-it from versioned copy instead of node_modules

### LINKED ISSUE #2376: tag-it uses an obsolete jQuery method
@kimrutherford I've noticed that the experimental conditions picker is broken on the `webpack` branch, because tag-it.js uses the `andSelf` method from jQuery, which has been deprecated and removed in favour of `addBack` in jQuery 3.0 ([source](https://api.jquery.com/andSelf/)). There's a lot of other deprecated functions in the script, but this is the only breaking one.

Unfortunately, we're loading tag-it from npm, and there doesn't seem to be a straightforward way to patch the file prior to bundling it with webpack. The easiest fix I can think of is checking in a patched version of tag-it.js that we can use in the interim, but I'm open to other ideas.

In future it would be best to migrate away from tag-it to something more up-to-date or compatible with AngularJS, since tag-it hasn't been updated in over six years, judging by its repository. I could look at adapting the solution used for the strain picker (since that also uses a tag-like interface), but it would need some extra work to integrate with the autocomplete.

jseager7:
(Fixes #2376)

This pull request has webpack load tag-it from a locally versioned script file (root/static/tag-it.js) instead of loading the file from node_modules, so that we can apply patches to tag-it more easily. We were already using a custom resolver to load the unminified version of tag-it, so the changes are minimal.

Our versioned copy of tag-it already contains a bunch of fixes that still aren't in the version from npm (and there's no newer version on npm as far as I can see), so I think using the patched versioned copy is definitely a better solution.

Included in the PR is a fix for an obsolete jQuery method, plus an additional fix to allow auto-completion of focused entries using the keyboard, since it seems the original fix from commit 2f19a53a18d4e1dc561eb07c81c1d1054a7dc06d broke due to a difference in selector class names (see #2163).

@kimrutherford I've tested the changes and everything seems to work as it should. If you don't think you need to test it yourself, feel free to merge.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,6 @@
     "jquery": "^3.5.1",
     "jquery-form": "^4.3.0",
     "jquery-ui": "^1.12.1",
-    "pnotify": "^5.1.2",
-    "tag-it": "2.0.0"
+    "pnotify": "^5.1.2"
   }
 }
--- a/root/static/js/tag-it.js
+++ b/root/static/js/tag-it.js
@@ -121,7 +121,7 @@
                 this.options.singleFieldNode = this.element;
                 this.element.addClass('tagit-hidden-field');
             } else {
-                this.tagList = this.element.find('ul, ol').andSelf().last();
+                this.tagList = this.element.find('ul, ol').addBack().last();
             }
 
             this.tagInput = $('<input type="text" />').addClass('ui-widget-content');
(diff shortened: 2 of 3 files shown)


kimrutherford:
I had a quick test and it works for me.  Thanks!

Issue #2299, opened 2020-06-09
Add help text for experimental conditions in PHI-Canto


jseager7:
Requested by @CuzickA in https://github.com/PHI-base/phi-eco/issues/10#issue-634722641

> The following text or similar would be useful to add to the conditions selection page
'Please add experimental conditions, conditions are aspects of the experimental setup independent of the genotype of the species under study ie pathogen/host strain or construct. They include any third party species used as a experimental assay system (heterologous species).

This is the text that needs to be customised:

![image](https://user-images.githubusercontent.com/37659591/84126621-0fe76400-aa36-11ea-89b9-1fde6fd1e41f.png)

@kimrutherford This will probably involve adding another configuration property that defaults to the text that's already there. The solution could follow what we did for the custom text on the finish form:

```yaml
curs_config:
  finish_form:
    extra_text_gene_information: >
      <p>
        Examples include:
        <ul>
        <li> protein domain data or other protein sequence features
        <li> missing gene names (synonyms)
        <li> transcriptional regulation (of the wild-type allele - transcriptional changes due to mutations can be captured in Canto)
        <li> complementation data (whether the gene complements a homologue in a different organism or vice versa; complementation may be full, partial or absent)
        <li> gene product name updates
        </ul>
        Most of these annotation types will be handled in future releases of the
        curation tool.
      </p>
```
If you're happy to explain how to add new configuration properties, then I'm happy to handle this myself.

jseager7:
> The solution could follow what we did for the custom text on the finish form:

Although in this case we probably don't need any rich formatting, so the configuration value could just be a plain string in the manner of the help text for annotations:

```yaml
    help_text: >
      A protein modification is a covalent modification or other change that
      alters the measured molecular mass of a peptide or protein amino acid
      residue
```

jseager7:
@CuzickA Would you want the following help text added on each phenotype annotation type? Based on the wording, it seems like it's most relevant for pathogen phenotype and host phenotype annotations.

> Please add experimental conditions, conditions are aspects of the experimental setup independent of the genotype of the species under study i.e. pathogen/host strain or construct. They include any third party species used as a experimental assay system (heterologous species)

 If we want the same text on metagenotype phenotypes, we may want to amend the wording as follows:

> ...conditions are aspects of the experimental setup independent of the **metagenotype containing** the species under study i.e. pathogen/host strain or construct.

Note that experimental conditions don't apply to GO annotations.

ValWood:
[...]

PR #2395, opened 2021-01-26
Enable configuration of experimental conditions help text

jseager7:
(References #2299)

@kimrutherford Here's the changes that we made to allow configuration of the help text above the experimental conditions picker. I was planning to commit this straight away, but I was wondering if you wanted to move the default help text into `canto.yaml` rather than having it in the Mason template.


jseager7:
Forgot to mention that I've tested the configured help text locally and it all works fine.

kimrutherford:
> but I was wondering if you wanted to move the default help text into canto.yaml

Yep, that makes sense.  Are you happy to have a go at that?

jseager7:
> Yep, that makes sense. Are you happy to have a go at that?

I am, but unless you want to make this change regardless of what happens here, it might be better to wait, because the child issue for this PR (#2299) is in flux at the moment and I'm not sure if we'll end up making the changes here after all.

kimrutherford:
> but unless you want to make this change regardless of what happens here, it might be better to wait

kimrutherford:
> but unless you want to make this change regardless of what happens here, it might be better to wait

There's no rush so let's leave it for now.

jseager7:
I think the discussion has been settled on this issue now, but I was looking at the proposed solution again, and now I'm wondering if it's the best choice.

My proposed solution assumes that the help text is unique to each annotation type:

```pl
$conditions_help_text = $annotation_type_config->{conditions_help_text};
```

But the text that is in the template is already global to every annotation type (or every annotation type that has conditions, since the template controls whether or not the conditions picker is shown):

```html
<div ng-if="showConditions()" class="curs-evidence-conditions">
```

So it seems like it would make more sense to move the configuration property for the help text to somewhere independent of annotation type – otherwise every admin user of Canto (e.g. me) is going to have to set the same configuration value multiple times.

@kimrutherford The question is, if we decide to configure this text globally, where's the best place for its configuration? Based on the existing configuration in `canto.yaml`, the `curs_config` object seems to be the most appropriate, but there's also the `messages` object. Which of these would you prefer? Or would you prefer to add the configuration somewhere else?

kimrutherford:
>  the curs_config object seems to be the most appropriate, but there's also the messages object

Yep, sorry, it's not very consistent.  I think putting to default help text in `curs_config` makes the most sense.

jseager7:
> I think putting to default help text in `curs_config` makes the most sense.

I've done this now. The property is called `experimental_conditions_help_text` and it's a direct child of `curs_config` (didn't make sense to nest it under anything else).

I've also made sure that the annotation edit modal (`annotationEditDialogCtrl`) uses the configurable help text as well, seeing as I forgot to do that in the first commit.

@kimrutherford I'd appreciate some double-checking of the code in `annotationEditDialogCtrl`. I copied the existing conventions with handling promises and the code seems to work, but the existing code seems quite complex so there could be some subtleties I've missed.

If everything looks fine, then this should be ready to merge.

kimrutherford:
> If everything looks fine, then this should be ready to merge.

Thanks James.  It all looks good including the changes in annotationEditDialogCtrl.

PR #2417, opened 2021-02-09
Use flex display mode for Metagenotype Management page

### LINKED ISSUE #2409: Host genotypes overlay  pathogen genotypes in metagenotype management if both alleles are long
An additional consequence  of wide pathogen and host genotype columns is this bug:

overlay

might be alleviated by
https://github.com/pombase/canto/issues/2408

jseager7:
Fixes #2409

This pull request changes the container for the genotype pickers on the Metagenotype Management page so that their layout is controlled with flexbox rather than Bootstrap column rules. This seems to stop one column overflowing under the other when the table gets too wide, as seen here:

overlay

Under the new layout system, the genotype picker columns will use use roughly half of the available horizontal space each. If the pathogen table grows wider than half the space, the host column shrinks (horizontally) to compensate. If the host column runs out of space for the host table, the whole column wraps below the pathogen column. This should keep both tables readable even in extreme cases.

@ValWood I'm not sure if you prefer to have the columns wrap under each other when the tables get too wide to fit on the screen. It's not ideal, but I hope that more vertical scrolling will be less annoying than horizontal scrolling.

- - -

@kimrutherford Note that as a consequence of the flex rules (mainly `flex-basis`), the pathogen table doesn't shrink to accommodate the host table; if the host table gets too wide, all it can do is wrap under the pathogen column. Trying to use `flex-grow: 1` on both columns led to behaviour that wasn't consistent with the previous layout, since the host column would shift around whenever the pathogen column changed size.

I'm not too keen on mixing Bootstrap columns with flexbox layouts (it would probably be better to choose one or the other) but I decided not to change everything in one go in favour of a faster fix. The only stuff that is still sized with Bootstrap columns seems to be max width anyway (`col-md-12`), so we might be able to use flexbox for everything on this page.

### DIFF
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1727,6 +1727,17 @@ a:not([href]) {
   font-weight: 700;
   margin-bottom: 10px;
 }
+.curs-metagenotype-picker-columns {
+  display: flex;
+  flex-wrap: wrap;
+}
+.curs-metagenotype-picker-column-pathogen {
+  flex-basis: 530px;
+  padding-right: 30px;
+}
+.curs-metagenotype-picker-column-host {
+  flex-shrink: 1;
+}
 
 .curs-warning-container {
   display: flex;
--- a/root/static/css/style.css
+++ b/root/static/css/style.css
@@ -1600,11 +1600,6 @@ a:not([href]) {
   margin-top: 10px;
 }
 
-.curs-metagenotype-manage-pathogen, .curs-metagenotype-manage-host {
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks for implementing that James.  I agree it's the best option.

> The only stuff that is still sized with Bootstrap columns seems to be max width anyway (col-md-12), so we might be able to use flexbox for everything on this page.

That would be good.  I'd be happy to reduce the dependency on Bootstrap.

Issue #2411, opened 2021-02-06
use  species filter to reduce metagenotypes displayed?


ValWood:
It might be useful if the organism filter only displayed the metagenotypes for the selected. species (perhaps with an 'All species' option to see everything.

Any ways to reduce the amount of information presented will be useful.  We often want to locate existing similar genotypes to duplicate them and after a while it becomes a lot to scan through.

![image](https://user-images.githubusercontent.com/7359272/107125548-96449600-68a2-11eb-99fe-5bb437ad21ff.png)

@CuzickA would that be useful?

jseager7:
This sounds like a good idea, especially considering that the Genotype Management page only shows the genotypes for the selected organism.

One other thing we might want to consider is additional filters by the pathogen or host side, in case curators want to see all metagenotypes related to a particular pathogen (on all hosts) or vice versa.

jseager7:
I've applied this change to both servers, so it's available for testing by curators.

Under the new system, all metagenotypes are shown when the page is first loaded, and stay shown until _both_ a pathogen and host are selected. This still gives curators the option to see all metagenotypes at once, as in the old system. Unfortunately there's no way to view all metagenotypes after selecting a pathogen and a host, because the organisms can't be deselected. The current solution is also a bit inconsistent with the Genotype Management page, because that doesn't default to showing all genotypes; it just shows no genotypes until an organism is selected.

We might need further discussion to decide on the following points:

* Do we want all metagenotypes shown when the page is first loaded, or should metagenotypes be hidden until a pathogen and a host are selected? (Note that only a filtered list of metagenotypes would be shown in this case.)

* If all metagenotypes are shown when the page is loaded, should we give the curator a way to later return to this view after selecting organisms? If so, how? There are multiple options:

  * Add a control that toggles the filtering, so that curators can switch between all metagenotypes and the filtered metagenotypes. The toggle won't persist between page loads, so hopefully it will be easy for curators to recover if they accidentally toggle the filter off (originally proposed by Val).

  * Add the ability to cancel the selection of the selected pathogen or selected host (or both at once), possibly by adding buttons next to the selector. (This could be awkward since the organism selector code is shared between the Genotype Management page and the Metagenotype Management page, but deselection wouldn't be useful for Genotype Management.)
  
[...]

PR #2419, opened 2021-02-11
Filter metagenotype list by selected organisms 

jseager7:
References #2411 

This pull request adds filtering to the table of metagenotypes on the Metagenotype Management page, such that it shows only the metagenotypes that match both the selected pathogen and the selected host. This is in keeping with the behaviour of the Genotype Management page, where only the genotypes for the selected organism are shown.

image

If the selected pathogen and host don't match any metagenotypes, a message is shown to the curator:

image

One thing that is different about this implementation is that _all_ metagenotypes are shown when the page is first loaded, and stay shown until both a pathogen and host are selected. I chose this just in case curators wanted a way to see all the annotations at once. Unfortunately there's no way to get back to this view after selecting a pathogen and a host, because the organisms can't be deselected. The original issue (#2411) suggested some control to disable the filters, but we need to decide on a specific design first.

- - -

As part of this change I moved almost all of the interaction with the `Metagenotype` service to the top of the directive hierarchy for the metagenotype management page, since I needed to access data that wasn't available except at the top of the hierarchy (the `metagenotypeManage` controller). I still don't really understand how the `Metagenotype` service works (it seems to be using different design patterns from our other services, namely event passing with `$rootScope`), but at least it's in a more useful location now.

@kimrutherford If you think everything looks okay, I'm happy for this to be merged now so we can get feedback from curators. I can open new pull requests for any follow-up issues.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -9260,7 +9260,7 @@ canto.directive('metagenotypeGenotypePicker',
 var metagenotypeListRow = function (CantoGlobals, Metagenotype, AnnotationTypeConfig) {
   return {
     scope: {
-      metagenotype: '=',
+      metagenotype: '<',
       showBackground: '<'
     },
     restrict: 'A',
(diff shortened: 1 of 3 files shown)


kimrutherford:
> Unfortunately there's no way to get back to this view after selecting a pathogen and a host, because the organisms can't be deselected.

I think that's fine for now.  The user can always reload the page or navigate away and then return.

> If you think everything looks okay, I'm happy for this to be merged now so we can get feedback from curators.

Yep!

I've done some testing.  Everything works and the changes all make sense.  Thanks!

Issue #2415, opened 2021-02-08
A suggestion: quick link to create WT allele


ValWood:
I wondered whether, since we use them so often, it might be useful to be able to create the WT alleles with a single click from the create genotype table (in the same way as deletion currently).
Having WT  more prominently might also make it clearer that expression levels for over-expression, or knockdown should be added to a WT allele

kimrutherford:
Would this be a PHI-Canto only button?

ValWood:
Yes, if we did it. 
I meant to tag @CuzickA  to see if useful.
It isn't super urgent. It's something that could be implemented if people get this wrong.  I think we saw some examples from the tester.

CuzickA:
Good idea. 
I think this would be really useful especially as we want curators to annotate all WT metagenotype controls.

jseager7:
> I think this would be really useful especially as we want curators to annotate all WT metagenotype controls.

What would be even better is a feature to automatically derive control metagenotypes from the mutant metagenotypes, but adding a WT button should be a lot easier than that.

jseager7:
@kimrutherford I'm happy to take a look at this if it's as simple as adapting the existing code in genotype_gene_list.html. One thing I might need to know is what configuration values to use for the wild type alleles. For reference, here's what we currently have for the deletion alleles in `GenotypeGeneListCtrl.makeDeletionAllele(geneId)`:

```js
var deletionAllele = {
  description: "",
  expression: "",
  gene_display_name: displayName,
  gene_id: geneId,
  gene_systematic_id: gene.primary_identifier,
  name: displayName + "delta",
  primary_identifier: "",
  type: "deletion",
};
```
One thing I'm a bit wary of is that there's a lot of code in `diploidConstructorDialogCtrl.ok()` that relates to wild type allele names. Is this a problem? I can't remember whether the quick allele buttons create diploids or haploids when Canto is in diploid mode. If they only create haploids then it should be unrelated to this change.

jseager7:
@ValWood I presume you'd still want the expression picker shown when the user uses the wild type shortcut button? Should the expression picker default to any particular value, if so? Currently there isn't any option selected by default, and if we want to make sure the curator pays attention to the expression level, then it's sensible to not have a default option.

Even with the shortcut, the user still has to select the strain and might have to select the expression level, so the shortcut won't be quite as fast as the deletion one (especially compared to PomBase where there's no strains).

ValWood:
[...]

PR #2421, opened 2021-02-23
Refactor GenotypeGeneList directive and controller

jseager7:
This PR has some preparatory refactoring to GenotypeGeneListCtrl that will hopefully make the changes required by issue #2415 easier to manage. The main change is extracting the controller from the directive definition and registering it with AngularJS separately (as I've done for other directives in the past). The other changes are mostly just code style and naming conventions.

@kimrutherford I haven't checked this in single organism mode, but it seems to work fine in pathogen-host mode (although note that pathogen-host mode doesn't exercise all of the code's functionality, such as disabling the deletion button when a deletion allele for a gene already exists). If you're happy to defer testing until the PR for #2415 is ready, then I'm fine with this being merged now.


kimrutherford:
I've done some testing in single organism mode and it all seems fine.  Disabling the deletion button works as expected too.

jseager7:
Thanks for testing. I'll merge this now.

PR #2423, opened 2021-02-25
Use Bootstrap UI tooltips for help icons

### LINKED ISSUE #2403: Replace jQuery UI dialogs with UI Bootstrap tooltips
Prompted by issue https://github.com/pombase/canto/issues/2397#issuecomment-769000771

Canto isn't consistent about the interface library that it uses for its help icon functionality. The help icons attached to section headings are using calls to `dialog()` (from jQuery UI), and don't link to any help documentation pages, while help icons elsewhere use UI Boostrap tooltips (through the `uib-tooltip` attribute), and those link to the relevant documentation pages on click. 

UI Bootstrap is more compatible with AngularJS, so we should probably migrate the dialogs to tooltips where possible. The only place this might be tricky is on the curation summary page, since AngularJS coverage might be lacking there.

jseager7:
(Fixes #2403)

I've changed all of the help icons to use UI Bootstrap tooltips instead of jQuery UI dialogs. I've tested the changes on the curation summary page (`front_gene_section.mhtml`) and the gene page (`gene_page.mhtml`), and they seem to work fine, although loading the `helpIcon` directive on the summary page seems to make Canto more laggy than before, based on my local copy.

@kimrutherford I can't find the pages in Canto that correspond to the following MHTML documents, so I can't test them.

* `root/curs/modules/interaction.mhtml`
* `root/curs/modules/interaction_evidence.mhtml`

I've switched them to Bootstrap UI tooltips anyway, but if these pages are unused, it might be better to remove the templates instead.

- - -

We might also want to update the technical documentation in `root\docs\md\canto_admin\configuration_file.md`, since the current text could be obsolete with regards to the new implementation:

> The keys under `help_text` identify a page in Canto and under the key is one or both of `inline` or `url`. The help text and link is rendered by the `inline_help.mhtml` template. If a `url` is given, the text under `inline` will be `title` attribute of a link with that URL. Without a `url` a help link (a "?" icon) will be shown and the `inline` text will be displayed in a pop-up DIV.

I'd propose the following (revisions in bold):

> The keys under `help_text` identify a **part of Canto's user interface;** under the key is one or both of `inline` or `url`. The help text and link **are** rendered **as a help icon with the text in a Bootstrap UI tooltip**. If a `url` is given, **clicking the icon will send the user to the page specified by the `url`.**

### DIFF
--- a/root/curs/front_gene_section.mhtml
+++ b/root/curs/front_gene_section.mhtml
@@ -15,7 +15,7 @@ Genes from this publication
 % } else {
 Annotate genes
 % }
-    <& /curs/inline_help.mhtml, key => 'choose_gene' &>
+    <help-icon key="choose_gene"></help-icon>
     </div>
   </div>
   <div class="curs-box-body">
@@ -66,7 +66,7 @@ Genotypes from this publication
 % } else {
 Annotate genotypes
 % }
-    <& /curs/inline_help.mhtml, key => 'choose_genotype' &>
+    <help-icon key="choose_genotype"></help-icon>
     </div>
   </div>
 %   if ($pathogen_host_mode) {
--- a/root/curs/gene_page.mhtml
+++ b/root/curs/gene_page.mhtml
@@ -14,7 +14,7 @@ var genes_in_session = <% $genes_in_session_js |n %>;
 <div class="curs-box curs-half-width-section">
   <div class="curs-box-title">
 Choose curation type for <% $gene->display_name() %>:
-    <& /curs/inline_help.mhtml, key => 'gene_page_choose_curation_type' &>
+    <help-icon key="gene_page_choose_curation_type"></help-icon>
   </div>
   <div class="curs-box-body">
   <ul class="annotation-start">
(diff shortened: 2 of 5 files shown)


kimrutherford:
> I can't find the pages in Canto that correspond to the following MHTML documents, so I can't test them.

Those pages for the long workflow for capturing interactions, which we aren't using anymore.  The only way to make an annotation now is to use the quick link.  I kept them around in case we changed our minds but I don't think that will happen.  I'll remove them after we merge this PR.

>I'd propose the following (revisions in bold):

I like the text but I think we can leave out the "Bootstrap UI" in this sentence since it's an implementation detail:

> The help text and link are rendered as a help icon with the text in a Bootstrap UI tooltip

jseager7:
@kimrutherford Thanks for the feedback. Here's the revised text. If you're happy with it then I'll add it to this PR.

> The keys under `help_text` identify a part of Canto's user interface **for which the help text should apply**. Under the key is one or both of `inline` or `url`. The help text and link are rendered as a help icon **with a tooltip that appears on mouse-over**. If a `url` is given, clicking the help icon will send the user to the page specified by the `url`.

kimrutherford:
> Here's the revised text. If you're happy with it then I'll add it to this PR.

That's great.  Thanks.

jseager7:
I've updated the documentation with the new text. Merging now.

PR #2424, opened 2021-03-01
Add shortcut buttons for wild type alleles

### LINKED ISSUE #2415: A suggestion: quick link to create WT allele

jseager7:
(References #2415)

This PR adds more shortcut buttons to the gene list on the Genotype Management page. The buttons allow creating a wild type allele for a given gene without having to specify the allele type.

image

After the button is clicked, a modal is shown with the Allele Type field hidden (the type is set to 'wild type' behind the scenes). The allele name, strain, and expression level can be set as normal.

image

- - -

@kimrutherford To implement this, I've added a flag to the configuration called `show_quick_wild_type_buttons` that controls whether the shortcut buttons are shown. I've also added an additional argument `alleleType` to the `makeAlleleEditInstance` function. When `alleleType` is set to any allele type (valid or not), the controller effectively locks the allele type field by setting its value to the specified allele type and then hiding the field with `ng-if`.

I chose this solution over creating a whole new dialog because **a)** it avoids re-implementing a lot of code and **b)** I thought the existing controller would take care of wild type allele naming conventions – although looking at the code, I'm not entirely sure if this is the case. Please let me know if you'd prefer a different approach (such as a dedicated modal), or if there's any extra handling needed.

Thankfully, JavaScript seems able to cope gracefully with functions that have undefined trailing arguments, so my change shouldn't break any existing function calls.

Some specific questions:

* Do you think that the `alleleType` argument name is obvious enough? I was considering renaming it to something more explicit, like `lockedAlleleType` or `lockToAlleleType`, so that it doesn't get confused with `alleleData.type`.

* Should `showAlleleTypeField` be a function instead of a variable? I've opted for a variable on `$scope` since I'm assuming that the value will only need to be evaluated once. Basically, the question reduces to whether or not `alleleData.type` can change after the modal is loaded. Here's the relevant code:

  ```js
  $scope.showAlleleTypeField = (
    ! args.alleleType && (
      $scope.alleleData.type != 'aberration' ||
      $scope.alleleData.type != 'aberration wild type'
    )
  );

### DIFF
--- a/root/curs/autohandler
+++ b/root/curs/autohandler
@@ -27,6 +27,7 @@ var diploid_mode = <% $diploid_mode ? 'true' : 'false' %>;
 var flybase_mode = <% $flybase_mode ? 'true' : 'false' %>;
 var max_term_name_select_count = <% $max_term_name_select_count %>;
 var show_quick_deletion_buttons = <% $show_quick_deletion_buttons %>;
+var show_quick_wild_type_buttons = <% $show_quick_wild_type_buttons %>;
 </script>
 
 <div id="disabled-overlay"></div>
@@ -84,4 +85,5 @@ my $diploid_mode = $c->config()->{diploid_mode} || 0;
 my $flybase_mode = $c->config()->{flybase_mode} || 0;
 my $max_term_name_select_count = $c->config()->{max_term_name_select_count} || 20;
 my $show_quick_deletion_buttons = $c->config()->{show_quick_deletion_buttons} || 0;
+my $show_quick_wild_type_buttons = $c->config()->{show_quick_wild_type_buttons} || 0;
 </%init>
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4802,6 +4802,11 @@ function GenotypeGeneListCtrl(
     });
   };
 
+  $scope.quickWildType = function (geneDisplayName, geneSystematicId, geneId) {
+    var alleleType = 'wild type';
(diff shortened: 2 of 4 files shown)


kimrutherford:
> I thought the existing controller would take care of wild type allele naming conventions – although looking at the code, I'm not entirely sure if this is the case.  Please let me know if you'd prefer a different approach (such as a dedicated modal), or if there's any extra handling needed.

I think it's all OK thanks to the change you made to extract the code into `updateAlleleType()` and then call it when setting the allele type.  But we can chat about that on Skype.

> Do you think that the alleleType argument name is obvious enough? I was considering renaming it to something more explicit, like lockedAlleleType or lockToAlleleType, so that it doesn't get confused with alleleData.type

I think that's a good idea.  `lockedAlleleType` sounds good to me.

I've done some testing in PomBase and FlyBase modes and everything worked fine.  Please go ahead and merge when you're ready.

Thanks for implementing that.

jseager7:
I've made the requested changes. I also fixed a small oversight where `lockedAlleleType` was hard-coded to 'wild type'. Since the allele type can be locked to any allele type, in future I could look at removing the special case for the deletion allele shortcut and having it reuse this new solution instead.

PR #2425, opened 2021-03-02
Allow configuration of gene confirmation page message 

jseager7:
Commit c142437a5c756eac24196424292cf9b1b0b8fc4a added a message to the top of `gene_list_edit.mhtml` in pathogen-host mode, explaining our remapping of NCBI Taxonomy IDs to species rank. At the time, I hard-coded the message into the template (and only enabled it in pathogen-host mode) because I didn't know any better. This PR extracts the message into a variable that is loaded from the configuration file.

The new configuration variable is called `gene_page_top_message`. Since the message is still only applicable to pathogen-host mode, I've added a placeholder to `canto.yaml` that is commented out by default.

Note that the substitution uses the `n` escape flag because the configuration value contains HTML markup (this was needed because the previous hard-coded message used rich formatting).

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -2362,6 +2362,9 @@ curs_config:
   follow_inverse_cv_names:
     - fission_yeast_phenotype
 
+  # A message for curators, shown at the top of the gene confirmation page.
+  # gene_page_top_message: SET_ME
+
   finish_form:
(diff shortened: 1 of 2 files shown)


kimrutherford (approved):
Thanks James.  That works for me locally.

Issue #2408, opened 2021-02-06
Hide redundant allele descriptions


ValWood:
I  realise we had older tickets about this issue, and I think we decided for now to ignore the problem. However, in PHI-Canto, it's very frustrating and really impede navigation between annotations and genotypes. 

These examples are full screen on my laptop.  Once you have a few screens of annotations to scroll through navigation becomes really difficult because you need to scroll right before you can scroll up or down.

Here are a couple of screenshots from the session I am working on at full screen resolution:

<img width="1665" alt="screen width" src="https://user-images.githubusercontent.com/7359272/107125148-0f8eb980-68a0-11eb-9ac1-28e5d300fe05.png">

<img width="1674" alt="screen width 2" src="https://user-images.githubusercontent.com/7359272/107125151-1289aa00-68a0-11eb-884f-bb8f8dbe6aad.png">

ValWood:
I'm going to suggest a small improvement that was probably suggested before.

There is quite a bit of duplication in  amino acid alteration names and descriptions

Gip1-G3-T324A,E325A,T326A,R327A,G329A,P330A(aaT324A,E325A,T326A,R327A,G329A,P330A)[Not assayed]

Gip1-G3-T324A,E325A,T326A,R327A,G329A,P330A

We made this change recently on PomBaseweb pages. It has the added benefit of making allele names and descriptions easier for a human to read
https://github.com/pombase/website/issues/1560

I would be in favour of implementing in Canto. Can anyone remember why we never did it? (we do this already for WT and deletion)

Could be displayed as

jseager7:
@ValWood So for the example above, and based on the discussion in https://github.com/pombase/website/issues/1560, would this be how you'd prefer to have the allele / genotype name displayed?

Gip1-G3-T324A,E325A,T326A,R327A,G329A,P330A (aa) [Not assayed]

Also, is the description part of this name automatically generated by Canto, or is it manually entered by the curator? I'm referring to the following text:

(aaT324A,E325A,T326A,R327A,G329A,P330A)

ValWood:
We enter it manually.
Sometimes the name  might be a diminutive. For example might give a mutant with 3 mutations to alanine AAA. 

Gip1-AAA(R327A,G329A,P330A)

But if we have curated the name it should be easy to know what the description is (and could also be visible by mouse over).

jseager7:
[...]

PR #2426, opened 2021-03-03
Remove redundancy from allele display names

kimrutherford:
Here's a possible fix for #2408

### DIFF
--- a/lib/Canto/Curs/Utils.pm
+++ b/lib/Canto/Curs/Utils.pm
@@ -1327,6 +1327,10 @@ sub make_allele_display_name
 
   $description ||= $type || 'unknown';
 
+  if ($name =~ /$description/) {
+    $description = '';
+  }
+
   if ($type =~ /^mutation|substitution/) {
     if ($type =~ /amino acid/) {
       $description =~ s/^/aa/g;
@@ -1337,7 +1341,8 @@ sub make_allele_display_name
     }
   }
 
-  if ($type eq 'other' && $name eq $description) {
+  if ($type eq 'other' && $name eq $description ||
+      length $description == 0) {
     return $name;
   }
 
--- a/lib/Canto/Curs/Utils.pm
+++ b/lib/Canto/Curs/Utils.pm
@@ -1327,11 +1327,11 @@ sub make_allele_display_name
 
   $description ||= $type || 'unknown';
 
-  if ($name =~ /$description/) {
+  if ($name =~ /[^a-z\d]$description$/) {
     $description = '';
   }
 
-  if ($type =~ /^mutation|substitution/) {
+  if ($type =~ /substitution/) {
     if ($type =~ /amino acid/) {
       $description =~ s/^/aa/g;
     } else {


jseager7 (changes requested):
I've tested the code now. I've suggested one change and asked one question. With my change applied, the code seems to work fine. Here's how it looks it looks in the different cases:

* Row 1: the name and description are different, and the description is shown as normal;
* Row 2: a disruption allele where the description (test) is a substring of the name (testing);
* Row 3: an amino acid substitution with identical name and description;
* Row 4: a nucleotide substitution with identical name and description.

image

Note that as part of testing, I realised that some existing code in `Utils.pm` doesn't seem to work as expected. I suspect it also conflicts with my proposed change.

```pl
  if ($type =~ /^mutation|substitution/) {
    if ($type =~ /amino acid/) {
      $description =~ s/^/aa/g;
    } else {
      if ($type =~ /nucleotide/) {
        $description =~ s/^/nt/g;
      }
    }
  }
```
There's a few oddities with this: currently, this regular expression is trying to match 'mutation' or 'substitution' at the start of the allele type string, but neither of those strings occur there in Canto's default allele types; it also won't match allele types containing 'insertion' or 'deletion'. Maybe it would be better to just remove this in favour of the new code.

Also, a few caveats about the proposed solution:

* Once the description is hidden, there's no way to see the allele description on the Genotype Management page. Arguably there's no need to show it, because the whole point is that the information is redundant. However, I think it might be surprising for users if their description suddenly disappears after creating their allele with no explanation of why it's vanished. I'm not sure of the best way to fix this. I thought about using an AngularJS filter to add a `title` attribute to the description placeholder (and adding an ellipsis in parentheses as a default placeholder instead of hiding the parentheses completely), but unfortunately there's already a Bootstrap UI tooltip on the rows:

  image

* It might not always be appropriate to hide the description, even if it's a substring of the name. I can think of a contrived example where 'cat' is a substring of 'catalysis', and while the characters are redundant, the _meaning_ of the substring absolutely isn't. Where the allele description follows a controlled vocabulary, this probably won't be a problem – but that may not apply for all allele types.

kimrutherford:
Hi James.  Thanks for reviewing the change.

> currently, this regular expression is trying to match 'mutation' or 'substitution' at the start of the allele type string,

There used to be some allele types that started with "mutation".  We changed those a while ago so that bit can be removed.
The RE is the same as `(^mutation)|(substitution)` so it's still matching the types with "substitution" as a sub-string.  I'm not sure that covers everything.  We'll need to check with Val and co. about that.

> but unfortunately there's already a Bootstrap UI tooltip on the rows

I think we could add the description to the tooltip.

> It might not always be appropriate to hide the description, even if it's a substring of the name.

We can probably tighten up the matching of the substring.  A lot of the pombe alleles look like "fan1-D651A(aaD651A)" so we could perhaps hide the description only if it's at the end of the allele name and proceeded by non letter or number like:

```perl
   if ($name =~ /[^a-z\d]$description$/) {
     $description = '';
   }
```

> Is it possible for $description to ever have zero length?

I think so.  For example if the type is "nonsense mutation" and the description is a substring of the name.  PomBase has alleles like:
```
type: nonsense_mutation
name: cwh43-K273->stop
description: K273->stop
```
In that case we'd be setting the description to "" so just returning the name would make sense.

kimrutherford:
@ValWood, @mah11, @CuzickA: we have code in Canto to display "aa" or "nt" at the start of allele description for some allele types.  So "fan1-q(R160E, R164E, K171E, R173E)" is displayed as "fan1-q(**aa**R160E, R164E, K171E, R173E)".

Currently that only happens for these allele types:
  - amino acid substitution(s)
  - amino acid insertion and substitution
  - nucleotide substitution(s)

Should it happen for any other types?  Here is the list of allele types that mention amino acid or nucleotide:

  - amino acid insertion
  - amino acid insertion and deletion
  - partial deletion and amino acid change
  - partial deletion, amino acid
  - nucleotide insertion
  - partial deletion, nucleotide

I'm particularly wondering about "partial deletion and amino acid change".

Thanks.

kimrutherford:
I've pushed some changes the hopefully get us closer to what's needed.

ValWood:
>currently, this regular expression is trying to match 'mutation' or 'substitution' at the start of the allele type string,

I believe all of these are long replaced.....if not we will fix

kimrutherford:
> > currently, this regular expression is trying to match 'mutation' or 'substitution' at the start of the allele type string,
>
> I believe all of these are long replaced.....if not we will fix

No problem.  There are no alleles with those types any more.

ValWood:
>Should it happen for any other types? Here is the list of allele types that mention amino acid or nucleotide:

I guess it should, but I have never noticed before that it doesn't. Mainly it's clutter because it isn't needed to interpret the allele, it is only really required if there is potential to confuse with nucleotide code...

jseager7:
> The RE is the same as (^mutation)|(substitution) so it's still matching the types with "substitution" as a sub-string. 

Ah, right. I'd always assumed that the pipe metacharacter created an implicit grouping on its neighbouring expressions, so the full expression would've been equivalent to `^(mutation|substitution)`. Thanks for clarifying that.

lib/Canto/Curs/Utils.pm
```diff
@@ -1327,11 +1327,11 @@ sub make_allele_display_name
 
   $description ||= $type || 'unknown';
 
-  if ($name =~ /$description/) {
+  if ($name =~ /[^a-z\d]$description$/) {
```
jseager7, line 1330: I'm not sure if this regular expression will match what was intended. It sounds like you wanted to match when the description "is at the end of the allele name and preceded by non letter or number", but this doesn't match numbers because the character class negation (`^`) applies to the _whole_ class. This is what the regex currently means:

image

Also, as far as I can tell, the expression is case sensitive, so it will still match descriptions preceded by an uppercase letter. I think this expression will do what you wanted:

```suggestion
  if ($name =~ /([^A-Za-z]|\d)$description$/) {
```

However, I'm not sure if regular expressions are the right solution here: using the example fan1-D651A(aaD651A), if I assume 'aaD651A' is the description (i.e. 'aa' is part of the description), then presumably this will never match because 'aaD651A' is not a suffix of 'fan1-D651A' – although without the 'aa', it would be. Maybe the match is supposed to be the other way round, but 'fan1-D651A' isn't a prefix of 'aaD651A' either, so I can't see how this will work. If the 'aa' prefix will be a problem, maybe there's a Perl function that can do suffix matching instead?

jseager7 (changes requested):

lib/Canto/Curs/Utils.pm
jseager7: > maybe there's a Perl function that can do suffix matching instead?

Actually I don't even think this would work because as I just said, "aaD651A is not a suffix". Maybe the simplest fix is just to remove 'aa' or 'nt' from the start of the description and then try to match the new string.

jseager7 (approved):

jseager7:
Marking my requested changes as resolved since they all turned out to be non-issues. This should be fine to merge now.

kimrutherford:
Thanks for the review James.

PR #2432, opened 2021-03-11
Prevent metagenotypes being deleted when used in extensions

### LINKED ISSUE #2427: JSON export error on undefined metagenotype identifier
I just tried to make a JSON export on our production server, and got the following error:
```
Can't call method "identifier" on an undefined value at lib/Canto/Curs/Serialise.pm line 291.
```
Looking at `Serialise.pm`, it looks like the error is due to a metagenotype ID being used in the range of an extension, but the ID has no corresponding metagenotype (probably because it's been deleted). 
```pl
        map {
          my $extension_bit = $_;

          if (defined $extension_bit->{rangeType}) {
            if (lc $extension_bit->{rangeType} eq 'metagenotype') {
              my $metagenotype =
                _get_metagenotype_by_id($schema, $extension_bit->{rangeValue});

              $extension_bit->{rangeValue} = $metagenotype->identifier();  # <-- error is here
            }
          }
        } @$extension_part;
```
The only extension we have that uses metagenotype IDs as its range is 'compared to control genotype' (compared_to_control).

@kimrutherford I suspect this is due to the problem we just discussed: Canto doesn't prevent a control metagenotype from being deleted if it is referenced by annotation extensions, meaning we end up with missing data.

I can think of two possible solutions to this:

1. ~~Write a quick script that I can use to track down the annotations that still have extensions referencing the nonexistent metagenotype, so that I can delete the extensions using Canto;~~ (done)

2. Amend the export script so that it can cope with nonexistent identifiers, and exports a placeholder value instead (maybe printing a warning at the same time).

jseager7:
(Fixes #2427)

I've had a go at fixing the bug that allowed metagenotypes to be deleted even when they were referenced in annotation extensions.

I've modified `Canto::Curs::GenotypeManager::delete_metagenotype` so that in addition to checking for any annotations that use the metagenotype, it also checks for any annotation extensions that use the metagenotype (specifically, extensions where the `rangeType` is 'Metagenotype' and the `rangeValue` is the ID of the selected metagenotype). If it finds any matching extensions, the method returns early with a relevant error message (similar to the existing implementation for annotations). I've also amended the error handling in the `Metagenotype` service on the front-end so that it shows a personalised error message when it receives the extension error.

One downside of the solution is that any annotation that has a matching extension usually doesn't belong to the metagenotype, so `delete_metagenotype` has to query all the annotations in the session. I adapted the code from `change-annotation-type-example.pl` to handle querying all extensions. The solution seems inefficient, but I can't see any way to improve it without some sort of caching when an extension referencing a metagenotype is created (and that means more complicated state management).

One optimisation that would make sense is to disable the check when `MetagenotypeID` isn't used as a range type in the annotation extension configuration, because then it's impossible for this situation to occur (so the check is redundant). If there's already a configuration flag for enabling Metagenotype Management (or metagenotype curation in general), that could be a reliable indicator, because having no metagenotypes also precludes the bug.

I've tested this manually on my local version and it seems to work fine: deleting a metagenotype is blocked when the metagenotype is referenced in an extension, and allowed otherwise. I haven't updated any unit tests to exercise this new functionality, mostly because I don't even know where the relevant test is.

@kimrutherford Could you review my changes to `delete_metagenotype` and let me know if there are any tests that need amending? Also whether we should bother making the optimisation of disabling the check when it's not needed?

### DIFF
--- a/lib/Canto/Controller/Curs.pm
+++ b/lib/Canto/Controller/Curs.pm
@@ -3189,6 +3189,11 @@ sub ws_annotation_create : Chained('top') PathPart('ws/annotation/create')
 
   my $json_data = $c->req()->body_data();
 
+  use Data::Dumper;
+$Data::Dumper::Maxdepth = 3;
+warn Dumper([$json_data]);
+
+
   $c->stash->{json_data} = $service_utils->create_annotation($json_data);
 
   $c->forward('View::JSON');
--- a/lib/Canto/Curs/GenotypeManager.pm
+++ b/lib/Canto/Curs/GenotypeManager.pm
@@ -768,6 +768,29 @@ sub delete_metagenotype
     return "metagenotype $metagenotype_id has annotations - delete failed";
   }
 
+  my $annotation_rs = $schema->resultset('Annotation');
+  while (defined (my $annotation = $annotation_rs->next())) {
+    my $extension = $annotation->data()->{extension};
+    if (defined $extension) {
+      map {
+        my $or_part = $_;
+        map {
+          my $and_part = $_;
+          my $is_metagenotype_extension = (
+            $and_part->{rangeType}
+            && $and_part->{rangeType} eq 'Metagenotype'
+          );
+          if ($is_metagenotype_extension) {
+            my $rangeValue = $and_part->{rangeValue};
+            if ($rangeValue == $metagenotype_id) {
+              return "metagenotype $metagenotype_id used in extensions - delete failed";
+            }
+          }
+        } @$or_part;
+      } @$extension;
+    }
+  }
+
   my $host_genotype = $metagenotype->host_genotype();
   my $pathogen_genotype = $metagenotype->pathogen_genotype();
 
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -757,10 +757,10 @@ canto.service('Metagenotype', function ($rootScope, $http, toaster, Curs) {
 
       }).catch(function (message) {
         if (message.match('metagenotype .* has annotations')) {
-          toaster.pop('warning', "couldn't delete the metagenotype: " +
+          toaster.pop('error', "couldn't delete the metagenotype: " +
             "delete the annotations that use it first");
         } else if (message.match('metagenotype .* used in extensions')) {
-              toaster.pop('warning', "couldn't delete the metagenotype: " +
+              toaster.pop('error', "couldn't delete the metagenotype: " +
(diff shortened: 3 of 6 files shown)


kimrutherford:
Thanks for implementing this.  It all looks good to me.

> The solution seems inefficient, but I can't see any way to improve it without some sort of caching when an extension referencing a metagenotype is created

Yep, it would be nice if it was easy to query this efficiently.  I'd like to improve things one day but for now I think it's fine.  Unless there are hundreds of annotations in a sessions the user isn't going to notice a delay.

> Could you review my changes to delete_metagenotype and let me know if there are any tests that need amending?

I've done some testing with no problems and the code changes look good.  Thanks!

> Also whether we should bother making the optimisation of disabling the check when it's not needed?

I think we should leave it for now.  I sure it will be fast enough as is.  This morning I needed to check all the extensions in all 5100 pombe Canto sessions and my little Perl script ran in 2 minutes for 82000 annotations.

kimrutherford:
> I haven't updated any unit tests to exercise this new functionality, mostly because I don't even know where the relevant test is.

I couldn't find a test for metagenotype deletion so I've added some here:
https://github.com/jseager7/canto/pull/4

It also tests deletion of a metagenotype that's referred to in an extension.

jseager7:
@kimrutherford It looks like the pull request has a lot of failing tests. I don't think it's related to my changes; I suspect it's a result of commit 1282ea12d11049c4dd3492ff6cd00c7dd985c69b, based on the fact that the tests that are failing are mostly the same as the tests changed by that commit.

kimrutherford:
I'll investigate that.  It's a bit weird as the tests are running fine on my desktop, using the same docker image.

I'll dig deeper tomorrow.

jseager7:
> I'll investigate that. It's a bit weird as the tests are running fine on my desktop, using the same docker image.

The tests also weren't failing on PR #2436. I tried re-running the job on this PR but it didn't change anything. Weird.

kimrutherford:
> I tried re-running the job on this PR but it didn't change anything. Weird.

I can't work it out.  I've double checked that I have exactly the same checkout with the same hash.  The tests work fine for me.  I think we should merge the branch and see if the tests pass.

jseager7:
That's fine with me. I'll merge now.

kimrutherford:
> I'll merge now.

Thanks.  Now I'm getting test failures locally so I'll be able to fix things.

kimrutherford:
I fixed the tests so it's all fine again.

PR #2436, opened 2021-03-15
Show genotype background for metagenotypes in feature chooser

### LINKED ISSUE #2435: Genotype background not shown in metagenotype picker
We noticed  whilst doing  a copy edit and adding background 
that the "background did not display in the  dropdown for PHI-Canto (I think it does for PomBase Canto?)

jseager7:
(Fixes #2435)

I've made some changes to `featureChooserFilter` so that it's able to display the genotype background for both genotypes in a metagenotype. Here's an example of the new system:

image

I've extracted all of the old code into functions so that it can be reused when formatting metagenotypes, since the metagenotype formatting in the new system effectively reduces to formatting two genotypes and concatenating them. The metagenotype display name is now effectively redundant in the metagenotype feature chooser, but that's not really a problem considering it's still used elsewhere (like annotation table rows).

Due to the genotype and metagenotype formatting sharing code, the genotype display name no longer prefixes its strain name with 'strain: '. The previous formatting wasn't consistent with how strains were displayed metagenotypes or feature tables. It's still possible to distinguish the background from the strain name because the background uses the 'bkg: ' prefix.

I've tested the new code under various feature types (including genes, even though I'm not sure if this filter is used for genes) and with certain properties set or undefined, and it seems to work well. The only problem is the fact that the logic for displaying organism names doesn't work as expected under the new system: as you can see from the image above, the organism names aren't showing here because the calling code doesn't specify that they should be shown, and the code only checks for `gene_id` and `genotype_id` properties when deciding whether to show the organism name. I'm marking this pull request as a draft until I've resolved the organism display issue.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -1696,7 +1696,7 @@ function featureChooserControlHelper($scope, $uibModal, CursGeneList,
 
 
 var multiFeatureChooser =
-  function ($uibModal, CursGeneList, CursGenotypeList, Curs, toaster) {
+  function ($uibModal, CantoGlobals, CursGeneList, CursGenotypeList, Curs, toaster) {
     return {
       scope: {
         features: '=',
@@ -1708,6 +1708,8 @@ var multiFeatureChooser =
       controller: function ($scope) {
         featureChooserControlHelper($scope, $uibModal, CursGeneList,
           CursGenotypeList, Curs, toaster);
+
+        $scope.showOrganism = CantoGlobals.multi_organism_mode;
 
         $scope.toggleSelection = function toggleSelection(featureId) {
           var idx = $scope.selectedFeatureIds.indexOf(featureId);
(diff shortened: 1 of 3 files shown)


jseager7:
Something I'm not sure about is if it ever makes sense to omit the organism names for metagenotypes. The only case I can think of where the organism names would be redundant is when both sides of the metagenotype are always the same species, or when Canto is configured to only ever have two species (e.g. one pathogen and one host). I'm not sure if the former case applies to FlyBase.

ValWood:
I don't think Flybase are using metagenotypes currently  (or if they plan to)

kimrutherford:
> I'm not sure if the former case applies to FlyBase.

In FlyCanto there is only ever one species and as Val says they aren't using metagenotypes.

jseager7:
I've fixed the organism display issues. `featureChooserFilter` now only uses the `showOrganism` argument to decide whether to show or hide the organism name. The argument is set by controllers that use the filter. If the argument is undefined, the organism names are hidden by default. This was a minor change since `featureChooserFilter` is only used by two directives.

Note that `multiFeatureChooser` had to be fixed due to a shortsighted past decision by me: the template was hard-coding the argument value as `true`, which under the new system would've led to the organism name being shown in the Transfer window under all organism modes. I suspect this would've affected FlyBase: if I remember correctly, they do use the Transfer annotation feature.

jseager7:
All the tests passed, so I'm merging now.

Issue #2428, opened 2021-03-04
Slim down the conditions and term suggestion sections of the edit dialog


kimrutherford:
For example:
![suggestion-suggestion-1](https://user-images.githubusercontent.com/90474/109954930-c585e000-7d46-11eb-88fa-ef7043367dbe.png)

ValWood:
<img width="1078" alt="Screenshot 2021-03-04 at 10 52 55" src="https://user-images.githubusercontent.com/7359272/109966262-f4b93480-7ce7-11eb-9f8f-9733200c02ac.png">


1. We have a box that says "type a condition" so do we really need to say "please add experimental conditions" ain't that bleeding obvious?
2. Conditions are yada yada (this is important but it could be under an info mouse over link)
3. Previously used conditions (click to add to list)
Maybe 
Previously used [+ ergwerg] [ + high temperature]

is sufficient?

kimrutherford:
How's this?:
![conditions-shrink-1](https://user-images.githubusercontent.com/90474/111426675-dd536000-8759-11eb-99f7-64f1b22009a8.png)

(The help text shows on mouse-over)

mah11:
fine

jseager7:
@ValWood For PHI-Canto, would you still want all the help text moved into the help pop-up? There was a lot of discussion in the past about stressing the fact that curators should add the delivery mechanism as an experimental condition, and it might not be as explicit if it's hidden in the help pop-up. Although, this might not matter if we've already told the curator that the delivery mechanism will be added as a condition at the start of the curation session.

ValWood:
There is a ticket here about the text that needs to be customised for PHI-Canto
https://github.com/pombase/canto/issues/2299
So, you are correct  in that you do require additional prompting.
I suggest for the PHI-Canto configuration  "start typing a condition, or delivery mechanism", and whatever  additions (if any) required for the pop up. Most of the  help about delivery mechanisms will be further forward  to prevent people from adding the organisms and strains used as  experimental tools but are not the species of interest.
The configuration  can be addressed in the other ticket as it is additional to the simplification described here.

jseager7:
[...]

PR #2438, opened 2021-03-18
Allow optional help text or tooltip for conditions

kimrutherford:
This conditions picker will now show help text at the top if the `experimental_conditions_help_text` config is set.  If `experimental_conditions_help_tooltip_text` is set then the configured text is show in a tooltip.  

Refs #2428

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -2381,8 +2381,11 @@ curs_config:
     - fission_yeast_phenotype
 
   # Help text shown above the experimental conditions picker
-  experimental_conditions_help_text: >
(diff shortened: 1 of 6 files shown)


jseager7:
@kimrutherford I've opened a pull request on this pull request at https://github.com/pombase/canto/pull/2439. I could've just pushed to this branch but I figured it would make reviewing (or rejecting) the changes easier if I kept things separate.

kimrutherford:
Thanks for your help James.  I think this is OK for merging now.  We can tweak later if needed.

PR #2439, opened 2021-03-18
Make conditions picker compatible with help icon directive 

jseager7:
This PR makes some more changes to the experimental conditions picker, on top of what's been done in https://github.com/pombase/canto/pull/2438.

The most significant change is replacing the existing implementation of the help icon with the `helpIcon` directive, since the previous solution just seemed to be re-implementing existing code. I moved the help icon text from `curs_config` to `help_text` in `canto.yaml`, so now this is the only markup that's needed:

```html
<help-icon key="experimental_conditions_definition"></help-icon>
```

As part of those changes, I had to rework the CSS layout for the conditions picker:

* I switched to a layout based on `float` because that was the only one that seemed to cooperate with the styling of the help icon and tag-it (`display: inline-block` just broke the layout). I looked at using `display: flex` but I was struggling to make anything work with that, so I just went with what was already working. The conditions picker has a fixed width now (close to what it used to have), the help icon has enough space to render its tooltip in the correct place, and the 'previous conditions' section clears to the left so that it stays below the floated condition picker input.

* I removed the Bootstrap row and column style classes from the conditions picker, because they were interfering with the float-based layout. As far as I could tell, the conditions picker isn't used in isolation, so it should probably be inheriting its layout from other components anyway.

I've also implemented a request from Val about shortening the 'Previously used conditions (click to add to list)' to 'Previously used conditions (click to add)'. We could shorten this further to 'Previously used', because the plus symbols on the condition buttons might be enough for the curator to infer that clicking the button will add the condition.

Here's how it looks in PHI-Canto, where we'll be using both help text features:

slim_condition_picker

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -2242,6 +2242,10 @@ help_text:
   genetic_interaction_definition:
     inline: Annotate interaction type and interacting genes.
     docs_path: "genetic_interaction_annotation"
+  experimental_conditions_definition:
+    inline: >
+      Conditions are the aspects of the experimental setup that are independent
+      of what cells (strain, genotype, constructs, etc.) are used.
 
 contact_email:
   address: "set_me@localhost"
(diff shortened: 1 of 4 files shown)


kimrutherford:
That looks great.  Thanks James.

PR #2442, opened 2021-03-22
Add more features to metagenotype filter

jseager7:
(References #2411)

This PR adds the ability to filter the metagenotype list on the Metagenotype Management page by _either_ the selected pathogen or the selected host: for example, if a pathogen is selected but a host isn't, then the list will only show the metagenotypes that include that pathogen, and will include all hosts (and vice versa).

I've also added a button to `organismSelector` that will unset the selected organism. This is needed to allow the curator to cancel the filter (to see all metagenotypes again after selecting an organism). The button is only shown if the binding `allowUnset` is true (or truthy). Here's how it looks:

image

@kimrutherford Note that allowing the organism to be unset means that any callback passed to `organismSelected` where `allowUnset` is true will have to handle a null value. That wasn't required before, since the value wasn't expected to become null after initially being set. I've amended the callbacks in the `metagenotypeManage` controller to handle this, but not anywhere else. I think this is safe for now since no other controller has `allowUnset` set to true, but I'm happy to fix it everywhere if needed.

### DIFF
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -4637,7 +4637,8 @@ var organismSelector = function () {
       organismSelected: '&',
       organisms: '<',
       initialSelectionTaxonId: '@',
-      label: '@'
+      label: '@',
+      allowUnset: '<',
     },
     restrict: 'E',
     templateUrl: app_static_path + 'ng_templates/organism_selector.html',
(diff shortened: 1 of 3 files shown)


jseager7:
I'm going to wait a bit for feedback in the linked issue before merging this.

PR #2447, opened 2021-03-31
Add table sections for haploid and diploid genotypes

### LINKED ISSUE #1941: Split genotype tables into haploid and diploid sections
Right now, genotypes get sorted into three tables on the genotype management page. But there are actually four, not three, kinds of genotype:

- haploid single locus
- haploid multi-locus
- diploid single locus
- diploid multi-locus

I think it would be good to have one table for each of these. It does not make sense to group haploid and diploid multi-locus genotypes (as it's doing now), especially if we continue to have separate tables for single-locus haploids and diploids. (It would be equally logically consistent to lump everything, but way too confusing to use.)

Aside: Both single-locus tables should offer the "combine selected" option, and both haploid genotype tables should offer the "create diploid" option (per #2012).

jseager7:
(Fixes #1941)

This PR updates the classification logic for the genotype table so that it is able to properly partition genotypes by loci and ploidy. There is still one table each for single- and multi-locus genotypes, and each table is divided into two sections: haploid and diploid.

image

If a multi-locus genotype has at least one diploid allele, it is classified as a diploid genotype.

- - -

@kimrutherford The main thing that needs reviewing is the changes to the classification logic. In order to classify mixed-ploidy genotypes correctly, I found that I couldn't rely on the length of `genotype.alleles` to determine the number of loci in a genotype. Instead, my solution classifies a genotype as single-locus if all of the gene IDs of its alleles are equal; otherwise the genotype is multi-locus. Diploid alleles are still determined by the presence of the `diploid_name` property.

I still have a shortcut for when the genotype only has one allele (which I've taken to imply that the genotype must be single locus and, conversely, cannot be multi-locus), but this new solution might be a bit wasteful when checking multi-locus genotypes when diploid mode is disabled.

My new code makes use of the higher-order functions `Array.prototype.some` and `Array.prototype.every` instead of `for` loops. These are apparently supported by IE9 (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some#browser_compatibility) and [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every#browser_compatibility)), and I think they make the code a lot more readable. For example, to check for single-locus genotypes, the code is simply:

```js
return genotype.alleles.every(geneIdEquals(firstGeneId));
```

I've also included checks for wild type genotypes (which I think still have the `alleles` property but with a length of zero), mostly so the classifier functions can be used anywhere in the code with less risk of exceptions, but if you think this is overly paranoid then I'm happy to remove them.

Note that adding the section header rows causes a minor graphical bug, where the background colour sequence of the rows doesn't reset at the start of each section. It should be possible to fix this by delegating the background colour setting to the `genotype-list-row` directive, but it needn't stop this PR being merged.


jseager7:
With regards to the out-of-order alternating row styling, it might be better to wait until #2448 is resolved before trying to fix that.

jseager7:
@kimrutherford I've merged your proposed changes, so if you don't think anything else needs changing, feel free to merge this PR.

kimrutherford:
Thanks James.  It looks good to me.

PR #2454, opened 2021-04-14
Add section for metagenotype annotation links on summary page

### LINKED ISSUE #2452: Add 'Annotate metagenotypes' section to Curation Summary page
In pathogen-host mode, the Metagenotype Management link is included under the 'Annotate genotypes' section, which isn't accurate since the page is about annotating metagenotypes, not genotypes.

image

It would be better if there were a dedicated section for the metagenotype annotation process, for example:

image

jseager7:
Fixes #2452

I've added a dedicated heading and section for the link to the Metagenotype Management page on the Curation Summary page. Previously, this link was part of the 'Annotate genotypes' section, which didn't make sense. I've also tried to clean up the markup a bit by making the Mason markup follow the indentation of the HTML, and I've added some extra containing elements so that the layout of the sections is consistent (see below for more details).

The diff is a bit chaotic due to all the indentation changes, so it's simpler to just show an image of the changes:

image

- - -

### Markup changes

In terms of the markup changes, I added a `curs-box` around each section of the list, so that each of the four sections are in their own containers.  For reference, the four sections (in order) are: 

1. Annotate genes
2. Delete or edit genes and organisms ('Edit gene list' in single organism mode)
3. Annotate genotypes
4. Annotate metagenotypes

Previously, the 'Annotate genotypes' section was actually nested in the container for 'Annotate genes', and because of a quirk in the styling rules, that meant it was indented further than the 'Annotate genes' section, giving the impression that it was somehow subordinate to that section when it shouldn't have been:

One side effect of the extra `curs-box-body` containers is that there's slightly more vertical spacing around the sections, but at least they're all using the same style rules now. I think most of the extra spacing is due to styling on the `feature-list-action` class, and I could adjust that if needed.

### Indentation and formatting

As previously mentioned, I've added indentation to the Mason markup in `front_gene_section.mhtml` so that it's easier to follow the structure of the HTML. The Mason still works fine as long as you don't indent the percent sign at the start of the line; unfortunately, that makes it very tedious to fix the indentation because you have to manually indent all the text past the first character. I'm not aware of any workaround for this, but I still think the benefits of readability are worth the extra effort. Here's an example of the markup with new indentation:

```
  <div class="curs-box-title">
    <div>
%     if ($read_only_curs) {
        Genes from this publication
%     } else {
        Annotate genes
%     }
      <help-icon key="choose_gene"></help-icon>
    </div>
  </div>
```

### DIFF
--- a/root/curs/front_gene_section.mhtml
+++ b/root/curs/front_gene_section.mhtml
@@ -3,108 +3,112 @@ $pub
 </%args>
 
 <script language="javascript" type="text/javascript">
-var organismsAndGenes = <% $organisms_and_genes_js |n %>;
+  var organismsAndGenes = <% $organisms_and_genes_js |n %>;
 </script>
 
 <div class="curs-box">
   <div class="curs-box-title">
     <div>
-
-% if ($read_only_curs) {
-Genes from this publication
-% } else {
-Annotate genes
-% }
-    <help-icon key="choose_gene"></help-icon>
+%     if ($read_only_curs) {
+        Genes from this publication
+%     } else {
+        Annotate genes
+%     }
+      <help-icon key="choose_gene"></help-icon>
     </div>
   </div>
   <div class="curs-box-body">
-% if ($pathogen_host_mode) {
-  <summary-page-gene-list> </summary-page-gene-list>
-% } else {
-% for my $org_and_gene_hashes (@gene_hashes_by_organism) {
-%   my @gene_hashes = @{$org_and_gene_hashes->{gene_hashes}};
-
-<div class="curs-front-page-gene-table">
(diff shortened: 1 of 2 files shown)


jseager7:
@kimrutherford Could you test this to make sure it looks okay in single organism mode? I've got conditionals in the markup that should hide the Annotate metagenotypes section unless Canto is in pathogen-host mode, but I haven't tested it.

kimrutherford:
> Could you test this to make sure it looks okay in single organism mode?

I've just tried it and it's all good.  Thanks.

kimrutherford (approved):
It all looks good to me.

jseager7:
Thanks for testing. Merging now.

Issue #2335, opened 2020-08-04
Improvements to 'with symbiont / host species' extension


jseager7:
Pull request https://github.com/pombase/canto/pull/2331 added support for choosing host or pathogen organisms in extensions. However, there's some limitations with the current solution that need fixing:

* Only show the 'with symbiont species' extension when the annotated organism is a host, and the 'with host species' extension when the annotated organism is a pathogen (strictly speaking, we need to check the organism related to the feature being annotated).

* Don't show the extension when there are less than two organisms of the respective type in the session. In the short term it could suffice to just show a warning message in the modal when this is the case, and prevent selection of the organism.

It was mentioned in https://github.com/pombase/canto/pull/2332#issuecomment-666679720 that the above solutions would require changes to `extensionConfFilter`.

jseager7:
I've looked at `extensionConfFilter` and it seems it has no reference to the feature being annotated, which we need in order to filter out irrelevant extensions in this case. This probably requires adding a new argument to `extensionConfFilter`.

The containing controllers will also need access to the feature so they can pass it to `extensionConfFilter`:

- `ontologyWorkflowCtrl` (for the step-based annotation workflow)
- `annotationEditDialogCtrl` (for the modal-based annotation workflow)
- `extensionBuilder` (the containing controller that is used for the extension list in both cases)

jseager7:
> In the short term it could suffice to just show a warning message in the modal when this is the case, and prevent selection of the organism.

I'm not sure if this is true, having tested it.

`extensionRelationEdit` can't distinguish between pathogens and hosts, so it always shows the message when there are fewer than two organisms, even when a pathogen gene annotation is being extended with a host, or vice versa (in which case it's perfectly fine to have just one organism of the opposite type). I think we need to solve the problem one level higher in the component hierarchy (as discussed above).

jseager7:
I've gotten close to fixing this by passing a new argument, `organismRole`, into `extensionConfFilter`, and using it to filter out extensions based on the role (pathogen or host) of the organism corresponding to the feature that's being annotated.

I haven't yet solved how to access the organism in all the controllers that need it. Possibly the best way would be to pass the whole feature, but some of the containing pages (e.g. ontology.mhtml) are Mason templates, and they only seem to support passing object properties as text to AngularJS components. Passing the feature would presumably need some form of serialisation.

[...]

PR #2460, opened 2021-04-27
Allow annotation extensions to be restricted by feature type

jseager7:
References #2335

In PHI-Canto, we have some annotation extensions for gene annotation types that should be hidden if the organism of the gene is a pathogen or a host. Currently, Canto can't restrict extensions in this way. I'm trying to change that by adding a new column to the extension configuration, called `feature_type`, that will restrict the extension to the specified feature type when set (otherwise, the restriction is ignored). This is similar to the restriction by annotation type added by PR #2367.

Unfortunately, even being able to restrict by feature type doesn't fully solve the problem, because the 'gene' feature type is too broad. What I really need is to restrict by the hypothetical feature types pathogen_gene and host_gene. I imagined these as some sort of feature subtype, analogous to the `feature_subtype` property in our `canto_deploy.yaml`:

```yaml
  - name: host_phenotype
    feature_type: 'genotype'
    feature_subtype: 'host'
```
But as far as I can tell, this doesn't exist on the actual feature objects provided by the server.

So far, I've gotten `ontologyWorkflowCtrl` working by adding the relevant feature types to the extension configuration table, then in order to generate the right feature type on the front-end, I fetched the organism data for the gene via the feature ID, and faked the correct feature type by passing a string value to `extensionBuilder`. The process of generating the feature type is shown below:

```js
        } else if (featureType == 'gene' && CantoGlobals.pathogen_host_mode) {
          CursGeneList
            .getGeneById(Number(featureId))
            .then(function (gene) {
              var organismRole = gene.organism.pathogen_or_host;
              $scope.finalFeatureType = organismRole + '_gene';
            });
```
This is already pretty convoluted, but the situation is even worse for the modal-based workflow. Firstly, there's a big stack of components to go through just to get to `extensionBuilder` (as shown below); and secondly, there are multiple entry points to the modal creation, meaning I'll probably have to compute the correct feature type in multiple locations. That might not be so bad if I extract the code into a global function, but it may mean having to inject more AngularJS services everywhere (in order to get the gene from a feature ID).

edit_modal_component_hierarchy

It would be ideal if the server could provide more details about the feature type (perhaps as a separate subtype property, so it doesn't risk breaking everything that relies on the more general types), but I can carry on trying to work around this on the front-end for now. Even if the feature type were provided by the server, there would still probably be a lot of work passing the data around (assuming the feature type is not there already).

@kimrutherford Can you think of a better solution to this problem?

- - -

Note that this PR also updates the tests relating to extension configuration, so now every row in `extension_config.tsv` has a correct value for `annotation_type_name` and `feature_type`. Previously, most of the `annotation_type_name` values were `undef`.

### DIFF
--- a/lib/Canto/Config/ExtensionConf.pm
+++ b/lib/Canto/Config/ExtensionConf.pm
@@ -70,7 +70,7 @@ sub parse {
       next if $line =~ /^#/;
 
       my ($domain, $subset_rel, $allowed_relation, $range, $display_text, $help_text,
-          $cardinality, $role, $annotation_type_name) =
+          $cardinality, $role, $annotation_type_name, $feature_type) =
             split (/\t/, $line);
 
       if ($domain =~ /^\s*domain/i) {
@@ -176,6 +176,7 @@ sub parse {
         cardinality => \@cardinality,
         role => $role,
         annotation_type_name => $annotation_type_name,
+        feature_type => $feature_type,
       );
 
       if ($domain =~ /(\S+)-(\S+)/) {
--- a/root/curs/modules/ontology.mhtml
+++ b/root/curs/modules/ontology.mhtml
@@ -58,7 +58,8 @@ Annotation extensions
                          feature-display-name='<% $feature_display_name %>'
                          term-id="{{currentTerm()}}"
                          is-valid="extensionBuilderIsValid"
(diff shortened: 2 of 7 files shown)


jseager7:
Today I learned that you shouldn't rename a branch that's linked to a pull request. This PR is replaced by #2461.

PR #2461, opened 2021-04-27
Allow annotation extensions to be restricted by feature type

jseager7:
References #2335; replaces #2460 due to the downstream branch being renamed.

In PHI-Canto, we have some annotation extensions for gene annotation types that should be hidden if the organism of the gene is a pathogen or a host. Currently, Canto can't restrict extensions in this way. I'm trying to change that by adding a new column to the extension configuration, called `feature_type`, that will restrict the extension to the specified feature type when set (otherwise, the restriction is ignored). This is similar to the restriction by annotation type added by PR #2367.

Unfortunately, even being able to restrict by feature type doesn't fully solve the problem, because the 'gene' feature type is too broad. What I really need is to restrict by the hypothetical feature types pathogen_gene and host_gene. I imagined these as some sort of feature subtype, analogous to the `feature_subtype` property in our `canto_deploy.yaml`:

```yaml
  - name: host_phenotype
    feature_type: 'genotype'
    feature_subtype: 'host'
```
But as far as I can tell, this doesn't exist on the actual feature objects provided by the server.

So far, I've gotten `ontologyWorkflowCtrl` working by adding the relevant feature types to the extension configuration table, then in order to generate the right feature type on the front-end, I fetched the organism data for the gene via the feature ID, and faked the correct feature type by passing a string value to `extensionBuilder`. The process of generating the feature type is shown below:

```js
        } else if (featureType == 'gene' && CantoGlobals.pathogen_host_mode) {
          CursGeneList
            .getGeneById(Number(featureId))
            .then(function (gene) {
              var organismRole = gene.organism.pathogen_or_host;
              $scope.finalFeatureType = organismRole + '_gene';
            });
```
This is already pretty convoluted, but the situation is even worse for the modal-based workflow. Firstly, there's a big stack of components to go through just to get to `extensionBuilder` (as shown below); and secondly, there are multiple entry points to the modal creation, meaning I'll probably have to compute the correct feature type in multiple locations. That might not be so bad if I extract the code into a global function, but it may mean having to inject more AngularJS services everywhere (in order to get the gene from a feature ID).

edit_modal_component_hierarchy

It would be ideal if the server could provide more details about the feature type (perhaps as a separate subtype property, so it doesn't risk breaking everything that relies on the more general types), but I can carry on trying to work around this on the front-end for now. Even if the feature type were provided by the server, there would still probably be a lot of work passing the data around (assuming the feature type is not there already).

@kimrutherford Can you think of a better solution to this problem?

- - -

Note that this PR also updates the tests relating to extension configuration, so now every row in `extension_config.tsv` has a correct value for `annotation_type_name` and `feature_type`. Previously, most of the `annotation_type_name` values were `undef`.

### DIFF
--- a/lib/Canto/Config/ExtensionConf.pm
+++ b/lib/Canto/Config/ExtensionConf.pm
@@ -70,7 +70,7 @@ sub parse {
       next if $line =~ /^#/;
 
       my ($domain, $subset_rel, $allowed_relation, $range, $display_text, $help_text,
-          $cardinality, $role, $annotation_type_name) =
+          $cardinality, $role, $annotation_type_name, $feature_type) =
             split (/\t/, $line);
 
       if ($domain =~ /^\s*domain/i) {
@@ -176,6 +176,7 @@ sub parse {
         cardinality => \@cardinality,
         role => $role,
         annotation_type_name => $annotation_type_name,
+        feature_type => $feature_type,
       );
 
       if ($domain =~ /(\S+)-(\S+)/) {
--- a/root/curs/modules/ontology.mhtml
+++ b/root/curs/modules/ontology.mhtml
@@ -58,7 +58,8 @@ Annotation extensions
                          feature-display-name='<% $feature_display_name %>'
                          term-id="{{currentTerm()}}"
                          is-valid="extensionBuilderIsValid"
-                         annotation-type-name="<% $annotation_type_name %>"></extension-builder>
+                         annotation-type-name="<% $annotation_type_name %>"
+                         feature-type="finalFeatureType"></extension-builder>
     </div>
   </div>
 
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -7244,15 +7244,23 @@ var annotationEditDialogCtrl =
           .then(function (data) {
             var extensionConfiguration = data[0];
             var termDetails = data[1];
-
             var subset_ids = termDetails.subset_ids;
+            var featureType = $scope.featureSubtype || $scope.annotation.feature_type;
 
-            if (extensionConfiguration.length > 0 &&
-                subset_ids && subset_ids.length > 0) {
-              $scope.matchingConfigurations =
-                extensionConfFilter(extensionConfiguration, subset_ids,
-                                    CantoGlobals.current_user_is_admin ? 'admin' : 'user',
-                                    $scope.annotationTypeName);
+            var hasExtensionsAndSubsets = (
+              extensionConfiguration.length > 0 &&
+              subset_ids &&
+              subset_ids.length > 0
+            );
+
+            if (hasExtensionsAndSubsets) {
+              $scope.matchingConfigurations = extensionConfFilter(
+                extensionConfiguration,
+                subset_ids,
+                CantoGlobals.current_user_is_admin ? 'admin' : 'user',
+                $scope.annotationTypeName,
+                featureType,
+              );
             } else {
               $scope.matchingConfigurations = [];
             }
--- a/root/static/js/canto-modules.js
+++ b/root/static/js/canto-modules.js
@@ -585,6 +585,21 @@ canto.service('CursGeneList', function ($q, Curs) {
 
     return q.promise;
   };
+
+  this.getGeneById = function (geneId) {
+    var genesPromise = this.geneList();
+    return genesPromise.then(function (genes) {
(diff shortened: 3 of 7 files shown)


jseager7:
Turns out fixing this in the modal workflow wasn't as bad as I thought, because the annotation edit dialog had pretty much all the data that I needed, and that code is shared by both component hierarchies.

@kimrutherford I'm marking this as ready to review. I've tested fairly thoroughly in pathogen-host mode, but I might need some help testing in PomBase mode and FlyBase mode. The code is restricted to only apply to pathogen-host mode, so I think the only thing that needs checking is that it doesn't have any unintended effects on extension filtering in other modes.

kimrutherford:
Hi James.  Sorry ignoring PHI-Canto for the last few weeks.

Thanks for working on this issue.  I've tried your changes locally and it all works well in PomBase mode.

Is this still relevant?:
> It would be ideal if the server could provide more details about the feature type

I'm happy to make server changes if that would help.

Do you think the following change in `annotationEditDialogCtrl` makes sense to match the rest of the PR?  The code in that controller is complicated but I think this is what's needed.

```diff
diff --git root/static/js/canto-modules.js root/static/js/canto-modules.js
index 23e295c25..8b2879bb9 100644
--- root/static/js/canto-modules.js
+++ root/static/js/canto-modules.js
@@ -7252,7 +7252,7 @@ var annotationEditDialogCtrl =
               $scope.matchingConfigurations =
                 extensionConfFilter(extensionConfiguration, subset_ids,
                                     CantoGlobals.current_user_is_admin ? 'admin' : 'user',
-                                    $scope.annotationTypeName);
+                                    $scope.annotationTypeName, $scope.chooseFeatureType);
             } else {
               $scope.matchingConfigurations = [];
             }
```

jseager7:
> Do you think the following change in annotationEditDialogCtrl makes sense to match the rest of the PR? 

Yeah, that was an oversight, thanks for pointing that out. I've handled the feature type here too. I've manually tested the changes and the extensions seem to work as expected when the term ID is changed.

- - -

> It would be ideal if the server could provide more details about the feature type
>> Is this still relevant?

**TL;DR:** Benefit versus cost is debatable, since the feature is very specific to PHI-base and probably won't see use elsewhere. Might be better to open a new issue or PR for these changes.

In terms of whether the server changes are still relevant: the main benefit to having the server provide the feature subtype is to simplify the front-end code and make the logic more extensible (for example, if we later wanted to add pathogen and host subtypes for genotypes, then it will be less wasteful to handle this on the back-end). However, at the moment, PHI-Canto is the only instance that needs this functionality, and it only needs it for one feature type (genes). So the trade-off might not be worth it unless the back-end fix is quick.

One caveat about the server solution: the server can't simply override the `feature_type` property, because then all the code that relies on that property stops working (e.g. a 'pathogen gene' is not a 'gene'). I think the safest fix is to include an additional `feature_subtype` property that contains the specific value that the extension filter needs, and the front-end will override the `feature_type` with the `feature_subtype` if it exists (which still leads to a bit of duplicated code handling this override everywhere, but it's no worse than the current solution).

Alternatively, we could loosen the logic for checking the feature types (e.g. _contains_ 'gene' instead of _is_ 'gene') but that's probably more work and more risky than what I just suggested.

If we add feature subtypes for genes, then it would probably make sense to provide feature subtypes for other features, where applicable. Genotypes and metagenotypes already have a `feature_subtype` property in Canto's configuration. Here's an example of the genotype `feature_subtype` being used to filter annotation types in `genotype_page.mhtml`:

```perl
    if ($pathogen_host_mode && $annotation_type->{feature_subtype}) {
      $annotation_type->{feature_subtype} eq $org_details->{pathogen_or_host};
    } else {
      1;
    }
```

If we had a feature subtype on the feature itself, then this check could be somewhat simplified:

```perl
    if ($pathogen_host_mode && $annotation_type->{feature_subtype}) {
      $annotation_type->{feature_subtype} eq $genotype->{feature_subtype};
    } else {
      1;
    }
```

Note that the feature subtype for genotypes is just 'pathogen' or 'host' as opposed to 'pathogen_genotype' or 'host_genotype'. If we go with the server solution, it might make more sense to qualify the feature subtype name so that it's less ambiguous when used in the 'feature type' column in the annotation extension configuration.

jseager7:
> I've just re-tested the current PR, including doing a reload of ontologies and extension files. Everything seems to be working well for pombe.

Thanks. I'll merge this now and open a new issue about the future server-side changes.

Issue #2463, opened 2021-05-03
recent canto change, restart approval


ValWood:
There is something a bit odd about this:

<img width="1274" alt="canto ticket" src="https://user-images.githubusercontent.com/7359272/116877111-68f35400-ac15-11eb-9745-6ea8b05e5196.png">

I *think* the restart approval button appears even if a session is not in approval. I recently used this to access a paper that was not yet submitted (I think), and now it is in approval, even though the authors did not yet curate it/submit it.

I tried again with one of my own sessions, and this is also now in approval.
I think I prefer to only be able to review, (or edit, but not necessarily in approval) from here?

mah11:
Yeah, I agree, this is kinda weird. I checked papers with a few different statuses, and it's using the same text and options for everything I tried except papers that have no session at all.

"Restart approval" makes sense only for papers with approved sessions (and not for sessions currently undergoing approval, still in progress, or created but nothing else done yet). I don't know if we care about the distinction between "currently being curated" and "has been curated" for the text.

Tangent: The link for "Admin only: publication page." is broken. It tries to go to `https://curation.pombase.org/......`, which of course isn't a thing. I think if we get that working, we wouldn't need a button for approval on the pombe "home" page.

kimrutherford:
It's not very recent.  It looks like it was changed in Jan last year: #2186

@jseager7 could you have a look at this?  Let me know if you'd like to chat about it.

ValWood:
I don't think it came into PomBase Canto until recently though...

jseager7:
> I don't think it came into PomBase Canto until recently though...

I think that's because PomBase's version of Canto is so far behind the latest version of the code (as stored on GitHub). Not sure though.

kimrutherford:
> I think that's because PomBase's version of Canto is so far behind the latest version of the code (as stored on GitHub).

That will be it.  I had forgotten.

jseager7:
This bug is my fault: I never added checks for the session state when I added the button. I'm not sure why the admin link is broken, but I'll take a look.

@kimrutherford Which session states should I be checking for? Based on [Midori's comment](https://github.com/pombase/canto/issues/2463#issuecomment-831896054) above, it sounds like I should be checking for `APPROVED` and `EXPORTED`, or maybe just `APPROVED`. Here's a list of the states from `State.pm`, for reference:

```perl
use constant {
  # user needs to confirm name and email address
  SESSION_CREATED => "SESSION_CREATED",
[...]

PR #2469, opened 2021-05-10
Fix 'Restart approval' button to respect state of curation session

jseager7:
References #2463

This PR should fix the remaining bugs with the publication search feature on Canto's homepage (the `pubmedIdStart` directive). The 'Restart approval' button now only shows if the curation session's state is 'APPROVED' (and if there is at least one session for the publication).

I also fixed a possible bug where the new code would try to run even if the publication data didn't load correctly.

@kimrutherford I've tested this locally and it seems to work fine. Feel free to merge if you don't notice any problems.


kimrutherford:
Works for me!  Thanks James.

PR #2501, opened 2021-11-22
Update generic email templates to better support PHI-Canto

jseager7:
See also: https://github.com/PHI-base/canto-docs/issues/1

This pull request updates the generic Canto email templates with some changes that we need for PHI-base (but which should be helpful for other databases too). Key changes include adding subsitutions for the curation application name, adding a signature lines to all the templates:
```
Sincerely yours,

The <% $config->{database_name} %> team
```
And rewriting the re-sent email template to be like a generic version of the PomBase template: previously, it just seemed to be a duplicate of another generic template.

- - -

Comments for @CuzickA:

* Note that in some cases (e.g. assigning sessions), the email is signed off with the name of the logged in user, rather than 'The PHI-base team'. If you'd prefer to keep the emails more anonymous, please let me know.

* You might want to check through the template for the reminder email for community curators who haven't started curating (`session_resent_body.mhtml`). I think there's a possibility it could come across as too assertive, especially considering that **a)** we're not well-known in the community like PomBase is, and **b)** we probably have nearly as much clout as they do. 

Comments for @kimrutherford:

* The PomBase templates can display  the name the curator is 'known as', where such an alias exists, using the following markup: `Dear <% $curator_known_as // $curator_name %>`. Should this change be added to the generic templates as well?

* I've used `<% $config->{name} %>` in place of 'Canto' so that the application name is substituted with 'PHI-Canto'. I assume that the `name` property in `canto_deploy.yaml` is always meant to be used as the name of the curation _application_, but the [Configuration settings](https://curation.pombase.org/pombe/docs/canto_admin/configuration_file) documentation isn't totally clear in this regard:

  > name
  > A one word name for the site. default: Canto

* Out of curiosity, is it really necessary to use the `Text::Wrap` formatter for the publication title? It looks a bit odd to have that wrapping when none of the other text does.

* If there's any changes you think shouldn't be in these generic templates (i.e. changes that are making the templates too specific) please let me know and I'll move the templates to their own versioned folder, like what was done for Japonicus.

### DIFF
--- a/root/email_templates/generic/session_resent_body.mhtml
+++ b/root/email_templates/generic/session_resent_body.mhtml
@@ -16,9 +16,10 @@ We recently invited you to contribute annotations based on your paper,
 to <% $config->{database_name} %>.
 
 We have noticed, however, that no action has been taken on the curation
-session. Please visit this page, where you can either start curating the
-paper, or reassign it to another author (e.g. the first author).
+session. Please visit the page linked below, where you can either start
+curating the paper, or reassign it to another author (e.g. the first author).
 
+Your curation link is:
   <% $session_link %>
 
 To reassign this session, use the "reassign" button, and enter the author's
--- a/root/email_templates/generic/session_resent_body.mhtml
+++ b/root/email_templates/generic/session_resent_body.mhtml
@@ -4,11 +4,12 @@ $publication_uniquename
 $publication_title
 $session_link
 $curator_name
+$curator_known_as
 $help_index
 $existing_annotation_count
 $logged_in_user
 </%args>
-Dear <% $curator_name %>,
+Dear <% $curator_known_as // $curator_name %>,
 
 We recently invited you to contribute annotations based on your paper,
 <% $publication_uniquename %> -
--- a/root/email_templates/generic/session_resent_body.mhtml
+++ b/root/email_templates/generic/session_resent_body.mhtml
@@ -40,7 +40,11 @@ up to date.
 
 Sincerely yours,
 
-<% $logged_in_user->name() %>
+% if ($config->{email_signature}) {
+<% $config->{email_signature} %>
+% } else {
+The <% $config->{database_name} %> team
+% }
 
 
 <%init>
--- a/root/email_templates/generic/session_resent_body.mhtml
+++ b/root/email_templates/generic/session_resent_body.mhtml
@@ -22,15 +22,14 @@ curating the paper, or reassign it to another author (e.g. the first author).
 Your curation link is:
   <% $session_link %>
 
+If you are unsure of how to curate this paper, or if you believe that this paper
+is not suitable for curation in <% $config->{name} %>, please send a message to
+<% $config->{contact_email}->{address} %> for assistance or advice.
+
 To reassign this session, use the "reassign" button, and enter the author's
 name and email address. To curate the publication yourself, click "start
 curating" and then follow the simple step-by-step instructions to capture the
 experimental data reported in your paper.
-
-If you begin curation, and are then unsure what to do next, or if you believe
-that this paper is not suitable for curation in <% $config->{name} %>, please
-send a message to <% $config->{contact_email}->{address} %> for assistance or
-advice.
 
 We greatly appreciate your contribution: all community annotations help keep
 <% $config->{database_name} %> literature curation comprehensive, accurate, and
--- a/root/email_templates/generic/session_resent_body.mhtml
+++ b/root/email_templates/generic/session_resent_body.mhtml
@@ -6,48 +6,40 @@ $session_link
 $curator_name
 $help_index
 $existing_annotation_count
+$logged_in_user
 </%args>
 Dear <% $curator_name %>,
 
-Congratulations on the recent publication of your paper:
-  <% $publication_uniquename %>
-<% wrap('  ', '  ', '"' . $publication_title . '"') %>
+We recently invited you to contribute annotations based on your paper,
+<% $publication_uniquename %> -
+<% wrap('  ', '  ', '"' . $publication_title . '"') %>,
+to <% $config->{database_name} %>.
 
-We invite you to contribute annotations based on this paper. We have
-identified the paper as suitable for curation in Canto, but if it
-reports data for a large number of genes, please let us know so we can
-collect the data by a different route.
+We have noticed, however, that no action has been taken on the curation
+session. Please visit this page, where you can either start curating the
+paper, or reassign it to another author (e.g. the first author).
 
-Your curation link is:
   <% $session_link %>
 
-When you follow the above link, you can start curating, or assign the session
-to a colleague (e.g. a member of your lab). You can annotate these data types
-in the tool:
+To reassign this session, use the "reassign" button, and enter the author's
+name and email address. To curate the publication yourself, click "start
+curating" and then follow the simple step-by-step instructions to capture the
+experimental data reported in your paper.
 
-<& ../annotation_type_list.mhtml, config => $config &>\
-% if ($existing_annotation_count > 0) {
+If you begin curation, and are then unsure what to do next, or if you believe
+that this paper is not suitable for curation in <% $config->{name} %>, please
+send a message to <% $config->{contact_email}->{address} %> for assistance or
+advice.
 
-For this paper, there are several previously curated annotations in the
-database. These existing annotations will be shown in the curation tool, but
-cannot be altered. If you notice any problems with existing annotations,
-please let the curators know.
-% }
-
-Help is available on each page, and you can also read
-the Canto manual (<% $help_index %>) or contact
-the help desk (<% $config->{contact_email}->{address} %>)
-for assistance at any time. You will also have an opportunity to provide any
-data that the tool does not capture so that curators can include it with your
-curation.
-
-The curation link will work indefinitely, so you can leave and return to the
-same URL as often as you want.
-
-We greatly appreciate your contribution.
-
+We greatly appreciate your contribution: all community annotations help keep
+<% $config->{database_name} %> literature curation comprehensive, accurate, and
+up to date.
 
 <% $session_link %>
+
+Sincerely yours,
+
+<% $logged_in_user->name() %>
 
 
 <%init>
--- a/root/docs/md/canto_admin/configuration_file.md
+++ b/root/docs/md/canto_admin/configuration_file.md
@@ -212,6 +212,11 @@ to the page specified by the `url`.
 This email address is shown anytime a contact address is needed. See the
 `contact.mhtml` template.
 
+### email_signature
+A custom signature line that is displayed at the end of the generic email
+templates. If not set, the signature will default to "The \[database_name\]
+team", where the database name is the value of the `database_name` setting.
+
 ### external_links
 Each external link configuration has three possible parameters:
 
--- a/root/email_templates/generic/session_accepted_body.mhtml
+++ b/root/email_templates/generic/session_accepted_body.mhtml
@@ -46,8 +46,11 @@ up to date.
 
 Sincerely yours,
 
+% if ($config->{email_signature}) {
+<% $config->{email_signature} %>
+% } else {
 The <% $config->{database_name} %> team
-
+% }
 
 <%init>
 use Text::Wrap;
--- a/root/email_templates/generic/session_accepted_body.mhtml
+++ b/root/email_templates/generic/session_accepted_body.mhtml
@@ -22,25 +22,31 @@ able to modify any annotations your colleague made.
 
 % if ($existing_annotation_count > 0) {
 For this paper, there are also several previously curated annotations in the
-database.  These existing annotations will be shown in the curation tool, but
+database. These existing annotations will be shown in the curation tool, but
 cannot be altered. If you notice any problems with existing annotations,
 please let the curators know.
 
 % }
-Help is available on each page, and you can also read the Canto manual
-(<% $help_index %>) or contact the help desk
+Help is available on each page, and you can also read the <% $config->{name} %>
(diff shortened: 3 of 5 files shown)


jseager7:
Also, I don't know if FlyBase uses these generic templates, but if they do, we should probably run the changes past them just to make sure that, for example, they're okay with addressing emails from 'The FlyBase team' and so on.

CuzickA:
Hi @jseager7
The changed files look good to me. I find it easier to read the actual email messages with the PHI-Canto / PHI-base text in, so I may have missed something here in these files.

I agree with your comment above about the reminder email being a bit too assertive. Perhaps we could soften it by adding text to the end of this section suggesting that if there are any queries please contact the curation email
image

Do we have an active link to PHI-base within the emails for people to click through to if they haven't heard of us before?

I would image that the automated emails could be signed off as 'The PHI-base team' and any follow up email discussions about the curated sessions could be signed off by the individual member of the PHI-base team dealing with the query. It would be useful to have some input from @ValWood and @kimrutherford about how the PomBase team keep track of curated papers and email correspondence. We currently have an email box for 'curation'. Would we try and create a folder for each paper/author in this email account? If all the 'curation' related emails are directed into this account it will be easier to keep track of correspondence and more accessible if we have multiple PHI-base team members interacting with community curators.

jseager7:
> I agree with your comment above about the reminder email being a bit too assertive. Perhaps we could soften it by adding text to the end of this section suggesting that if there are any queries please contact the curation email

We already have a reference to the curation email in a later paragraph, so I could rearrange the text such that the paragraph for the contact details comes immediately after the reminder, and before the instructions on how to reassign a session. For example: 
```
We have noticed, however, that no action has been taken on the curation
session. Please visit this page, where you can either start curating the
paper, or reassign it to another author (e.g. the first author).

...

If you begin curation, and are then unsure what to do next, or if you believe
that this paper is not suitable for curation in <% $config->{name} %>, please
send a message to <% $config->{contact_email}->{address} %> for assistance or
advice.
```
If we decide to do this, I'll reword the first part of the second paragraph as follows: 
```
If you are unsure of how to curate this paper, or if you believe
that this paper is not suitable for curation...
```
- - -
> Do we have an active link to PHI-base within the emails for people to click through to if they haven't heard of us before?

Not at the moment. In order to add this, I'd have to add an entry to the Canto configuration file with a URL for the PHI-base website in it, assuming that configuration doesn't already exist somewhere. That's a fairly quick change to make.
- - -
> Would we try and create a folder for each paper/author in this email account?

We could, but it might be simpler to sort by the From field or filter by PubMed identifiers in the main inbox. Using Conversation view in Outlook would help to keep emails from a curation session grouped together, assuming the author replies to the automated emails. We could look at Rules if we need anything more sophisticated.

CuzickA:
> > If we decide to do this, I'll reword the first part of the second paragraph as follows:
If you are unsure of how to curate this paper, or if you believe
that this paper is not suitable for curation...

This sounds good to me

> Do we have an active link to PHI-base within the emails for people to click through to if they haven't heard of us before?

> Not at the moment. In order to add this, I'd have to add an entry to the Canto configuration file with a URL for the PHI-base website in it, assuming that configuration doesn't already exist somewhere. That's a fairly quick change to make.

I think this would be really useful and also increase our PHI-base traffic ;-)

> Would we try and create a folder for each paper/author in this email account?

> We could, but it might be simpler to sort by the From field or filter by PubMed identifiers in the main inbox. Using Conversation view in Outlook would help to keep emails from a curation session grouped together, assuming the author replies to the automated emails. We could look at Rules if we need anything more sophisticated.

This may be a silly question but is PHI-Canto connected to our curation email box? If we assign papers to authors for curation via the PHI-Canto Admin pages will copies of all automated emails be stored in our curation email box? Is the author curation management predominantly done via the Admin pages or via our curation email?

ValWood:
At PomBase we link our curation e-mail responses, and Canto questions up to our helpdesk.

ValWood:
In Canto Admin interface you can see the history of how many e-mails have been sent. We don't store copies of these e-mails (no need because they are all  the same 2 e-mails, we just need to recorded how many  requests were sent).

I strongly advise that you sign of from a  Kim H-K . People are more likely to respond if the e-mail comes from an individual that they know.

CuzickA:
Thanks @ValWood. So automated emails are stored in the Admin pages and follow up emails would be stored in our curation inbox. 

I'll check with Kim H-K to see what she thinks about the sign off options.

ValWood:
The e-mails are not stored at all, just a count of how many reminders were sent. We know the e-mail content and the addressee so we don't need to store  the actual e-mail.

jseager7:
> This may be a silly question but is PHI-Canto connected to our curation email box? If we assign papers to authors for curation via the PHI-Canto Admin pages will copies of all automated emails be stored in our curation email box?

Not sure about the second part, but the PHI-Canto configuration file (`canto_deploy.yaml`) has fields for email addresses that we can specify. There are three address options under the `email` property. According to the comments in the default configuration file (`canto.yaml`), here's what the options do:

* `from_address`: used as the "from" field in all outgoing messages.
* `admin_address`: the contact address of the admin curators, used to alert about completed sessions and problems.
* `noreply_address`: there's no documentation for this, but I think it's used as a fallback if the `admin_address` is not set.

There's also one address under the `contact_email` property, simply called `address`, which apparently is "shown anytime a contact address is needed" ([source](https://curation.pombase.org/pombe/docs/canto_admin/configuration_file#contact_email)). Not sure exactly what that means. 

For PHI-Canto, all these options currently use the same curation at PHI-base email address.

jseager7:
> I strongly advise that you sign off from Kim H-K. People are more likely to respond if the e-mail comes from an individual that they know.

I agree, but just to note, the signature line in the generic email templates isn't currently configurable. In order to do this, I'd either have to add a configuration option for the signature text (something like `email_signature`), or create our own set of personalised email templates for PHI-base that are checked in to the repository.

ValWood:
I wonder how we do it ? @kimrutherford

root/email_templates/generic/session_resent_body.mhtml
```diff
@@ -6,48 +6,40 @@ $session_link
...
 
-We invite you to contribute annotations based on this paper. We have
-identified the paper as suitable for curation in Canto, but if it
-reports data for a large number of genes, please let us know so we can
-collect the data by a different route.
+We have noticed, however, that no action has been taken on the curation
+session. Please visit this page, where you can either start curating the
+paper, or reassign it to another author (e.g. the first author).
 
-Your curation link is:
```
kimrutherford, line 21: Can we keep this line or something like it?  I'd like to be clear what the link is for.

kimrutherford (changes requested):
It all looks good to me except for my one small comment.

kimrutherford:
> The PomBase templates can display the name the curator is 'known as', where such an alias exists, using the following markup: Dear <% $curator_known_as // $curator_name %>. Should this change be added to the generic templates as well?

Yep, it makes sense to add that.

>  I assume that the name property in canto_deploy.yaml is always meant to be used as the name of the curation application,

You're right, it is.  I've tweaked the documentation.

> Out of curiosity, is it really necessary to use the Text::Wrap formatter for the publication title? It looks a bit odd to have that wrapping when none of the other text does.

The title is the only interpolated value that's long.  Wrapping it made the email text long better in my email client.

> If there's any changes you think shouldn't be in these generic templates

I think the changes are OK.

> Also, I don't know if FlyBase uses these generic templates, but if they do, we should probably run the changes past them just to make sure that, for example, they're okay with addressing emails from 'The FlyBase team' and so on.

It shouldn't be a problem for them as the aren't going to use Canto for community curation.  It's internal only.

root/email_templates/generic/session_resent_body.mhtml
jseager7: I'll probably have to reword the previous paragraph as well but yes, I'll make that change.

jseager7:
> The title is the only interpolated value that's long. Wrapping it made the email text long better in my email client.

Maybe email clients vary in their text wrapping behaviour? In my case it's the wrapped text that looks out of place because it's so much shorter than the other lines. I guess the wrapping does improve readability somewhat, so I'm happy to leave it as-is.

> It shouldn't be a problem for them [FlyBase] as they aren't going to use Canto for community curation. It's internal only.

I thought that might've been the case. Thanks for confirming.

root/email_templates/generic/session_resent_body.mhtml
jseager7: Fixed by https://github.com/pombase/canto/pull/2501/commits/ee018c5f2c13cea36dc78462797210042cfa1114.

jseager7:
@kimrutherford I've made your one requested change about the session link explanation and I've added the `$curator_known_as` variable to all the relevant templates.

@CuzickA I've made your requested changes about softening the tone of the reminder email. I still need to add a link to the PHI-base website somewhere in the email. Do you have any preference for where this link should be placed in the body of the email?

With regards to the changes to the signature line, that might need some more thought. The fastest way to fix it would be to add specialised templates for PHI-base, but it would be good to have a configurable signature line generally available. I'm just not sure of the best way to do it.

jseager7:
@kimrutherford Seems the PomBase version of `session_reassigned_body.mhtml` doesn't use `$curator_known_as`, so the test `t/70_email_util.t` fails when I include it.

I tried to fix this by adding `$curator_known_as` as an argument in `reassigner_body.mhtml`, but that failed for a different reason:
```
# '
#     doesn't match '(?^:Dear Val Wood)'
```
FYI, the address line in the printout is the following:
```
# Dear Dr Val,
```
I've reverted that one use of `$curator_known_as` for now, but should the test and PomBase email template be updated now to fix this everywhere?

kimrutherford:
> but should the test and PomBase email template be updated now to fix this everywhere?

Hi James.

I've fixed that in the main branch: 7dc64a5f4be9c

kimrutherford:
> I agree, but just to note, the signature line in the generic email templates isn't currently configurable. In order to do this, I'd either have to add a configuration option for the signature text (something like email_signature),

I think that makes sense.

You could do something like this in the template so the `email_signature` is optional:

```perl

 Sincerely yours,

 % if ($config->{email_signature} {
 <% $config->{email_signature}
 % } else {
 The <% $config->{database_name} %> team
 % }
```

kimrutherford (approved):

CuzickA:
In today's meeting we agreed to sign off as

Kim Hammond-Kosack and the PHI-base team

jseager7:
> You could do something like this in the template so the email_signature is optional:

That sounds like exactly what we need, thanks.

jseager7:
> In today's meeting we agreed to sign off as
> Kim Hammond-Kosack and the PHI-base team

@CuzickA I was looking at the email templates while making this change, and signing off as Kim H-K might not be the best choice for the 'session reassigned' template. That's because the name and email of the assigner is already shown at the top of the email:
```
<% $reassigner_name %> <<% $reassigner_email %>> has invited you to contribute
annotations to <% $config->{database_name} %> based on this paper:
```
It could be confusing for the recipient if the person inviting them to contribute is different from the person who signed the email, so it might be better to just sign off as 'The PHI-base team' in this case.

ValWood:
The email  is generated by the system  and comes from "helpdesk" so I don't think this is an issue?

jseager7:
> The email is generated by the system and comes from "helpdesk" so I don't think this is an issue?

The actual sender of the email will presumably be the PHI-base helpdesk email address, but the signature line at the end of the email body will be from Kim H-K (and the PHI-base team) while the `<% $reassigner_name %>` and `<% $reassigner_email %>` placeholders at the start of the email body will be the name of the person originally assigned to the session that initiated the reassignment (which is probably going to be Alayne or some other professional curator). I just thought the combination seemed a bit incongruous.

ValWood:
ok i see.

CuzickA:
> > In today's meeting we agreed to sign off as
> > Kim Hammond-Kosack and the PHI-base team
> 
> @CuzickA I was looking at the email templates while making this change, and signing off as Kim H-K might not be the best choice for the 'session reassigned' template. That's because the name and email of the assigner is already shown at the top of the email:
> 
> ```
> <% $reassigner_name %> <<% $reassigner_email %>> has invited you to contribute
> annotations to <% $config->{database_name} %> based on this paper:
> ```
> 
> It could be confusing for the recipient if the person inviting them to contribute is different from the person who signed the email, so it might be better to just sign off as 'The PHI-base team' in this case.

@jseager7 I agree.

jseager7:
I've added the configurable signature line now.

I've just realised my changes mean that some of the templates default to signing off as `The <% $config->{database_name} %> team` instead of `<% $logged_in_user->name() %>` when the signature line isn't configured. Not sure if that's ideal.

@kimrutherford Does Canto provide a way to preview the output of the email templates after templating? Otherwise the only way I can think to test this locally is sending the email and intercepting it using external tools.

kimrutherford:
> Does Canto provide a way to preview the output of the email templates after templating?

No, sorry.  I usually test by assigning (or reassigning or whatever) a session to myself or to Val on our test server.

jseager7:
I've added some documentation for the `email_signature` property to the Configuration settings help page.

I'm going to have to merge this now in order to test it on our servers.

I would've updated `t/70_email_util.t` to test the signature line but the changes conflicted too much with the existing tests. Also, it seems like the test only tests the body of the PomBase templates, based on the following lines:

```perl
$config->{email}->{templates}->{session_assigned}->{body} =
  "email_templates/pombase/session_assigned_body.mhtml";
```

jseager7:
For the record, there were no problems with these updated email templates being sent once we started a new curation session. Reassigning an existing curation session didn't seem to trigger the emails but that might've been because I had been assigned to the session once before.

Issue #2505, opened 2021-12-08
Metagenotype changes should propagate to annotation extensions


jseager7:
(Requested by @CuzickA)

We've noticed that once a metagenotype is used in our **compared_to_control** annotation extension (which has `MetagenotypeID` as its range), the extension value doesn't update if the metagenotype it references is changed. This means the curator has to recreate all of the annotation extensions manually (sometimes upwards of 20 at a time).

@kimrutherford Am I right in thinking that the annotation extension merely sets its value to a copy of the display name of the metagenotype? If so, how difficult would it be (for me) to change the extension so that it references the current version of the metagenotype? I doubt this has ever been a problem before, because the only other feature that's used as a range ID is `GeneID`, and that's immutable in a curation session.

Another option could be to add a feature for bulk-editing annotation extension range values across multiple annotations, but that could be even more complicated than the other fix.

kimrutherford:
> Am I right in thinking that the annotation extension merely sets its value to a copy of the display name of the metagenotype?

It stores the metagenome database ID and the display name.  If the metagenotype changes the display name stored in the extension isn't updated.  The current implementation isn't great.  In the database the extension should probably just contain the ID and the Perl code should fill in the name before returning the JSON to Javascript code.  I'll have a look at doing that.

ValWood:
This is similar to needing to edit "multi-gene genotypes" when one of the contributing genotype changes. I thought there was a ticket about this, but I can't find it.

mah11:
> This is similar to needing to edit multi-~gene~locus genotypes when one of the contributing ~genotype~ allele names or descriptions changes. I thought there was a ticket about this, but I can't find it.

The main one is #2505 (and it links to #1901, but that one has less detail).

mah11:
corrected ticket link (#2304) ...doh

jseager7:
@kimrutherford This problem keeps showing up now that we have new curators on the project. Could you please take another look at this when you have time? I'm happy to help test or make changes.

kimrutherford:
Just to be clear, what sort of metagenotype changes are we talking about?  Can metagenotypes change after creation?

jseager7:
If the genotype display names are always updated correctly, perhaps it's easy for PHI-Canto to just concatenate the genotype display names to form the metagenotype display name at runtime, like you suggested.

[...]

PR #2557, opened 2022-04-08
Don't store the metagenotype display name in extensions

kimrutherford:
This PR changes the code to:
  - not store the rangeDisplayName for metagenotypes in extensions
  - fill in the rangeDisplayName for metagenotypes in extensions when annotation details are requested from the server
     - the Metagenotype ID in the the rangeValue field of the extensions is used to look up the metagenotype as needed

### DIFF
--- a/lib/Canto/Curs/Serialise.pm
+++ b/lib/Canto/Curs/Serialise.pm
@@ -40,6 +40,7 @@ the Free Software Foundation, either version 3 of the License, or
 use strict;
 use warnings;
 use Carp;
+use feature 'state';
 
 use JSON;
 use Clone qw(clone);
@@ -180,7 +181,8 @@ sub _get_annotations
 
   die "no schema" unless $schema;
 
-  my $organism_lookup = Canto::Track::get_adaptor($config, 'organism');
+  state $ontology_lookup = Canto::Track::get_adaptor($config, 'ontology');
+  state $organism_lookup = Canto::Track::get_adaptor($config, 'organism');
 
   my $rs = $schema->resultset('Annotation');
 
@@ -313,6 +315,15 @@ sub _get_annotations
                 _get_metagenotype_by_id($schema, $extension_bit->{rangeValue});
 
               $extension_bit->{rangeValue} = $metagenotype->identifier();
+
+              my %metagenotype_details =
+                Canto::Curs::Utils::make_metagenotype_details($schema, $metagenotype,
+                                                              $config,
+                                                              $ontology_lookup,
+                                                              $organism_lookup);
+
+              $extension_bit->{rangeDisplayName} =
+                $metagenotype_details{metagenotype_display_name};
             }
           }
         } @$extension_part;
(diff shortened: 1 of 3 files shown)


jseager7 (changes requested):
@kimrutherford The changes to the user interface seem to work fine for me on my local copy. I've tested renaming a metagenotype both in existing sessions and in new sessions and I can't see any problems there.

The only problem is that the metagenotype names aren't updated in the JSON export, so the code probably needs changing there to mirror the changes in `Utils.pm`. For example, I renamed the allele in the following JSON from `lasI+` to `lasI_test`:

```json
"P33883:11fa7733063e791c-3" : {
    "allele_type" : "wild_type",
    "gene" : "Pseudomonas aeruginosa P33883",
    "name" : "lasI_test",
    "primary_identifier" : "P33883:11fa7733063e791c-3",
    "synonyms" : []
},
```
But in the exported annotation extensions, it still has the old name:
```json
{
    "rangeDisplayName" : "lasI+[WT product] Pseudomonas aeruginosa (PA14) / wild type Caenorhabditis elegans (N2)",
    "rangeType" : "Metagenotype",
    "rangeValue" : "11fa7733063e791c-metagenotype-20",
    "relation" : "compared_to_control"
}
```

- - -

With regards to the potential performance issue mentioned in https://github.com/pombase/canto/issues/2505#issuecomment-1092492826, I tested the page load time on a session with 76 annotations and it seemed to load _faster_ than before, compared to commit 13cbc4d6d91efaf15e60d49cd2c4b13a1204b639 on the `master` branch: loading the page before the changes takes on average about 13 seconds, whereas after the changes it averages about 8 seconds.

I only tested using the debugging tools in Firefox, so I don't know how reliable these numbers really are. It's possible that some other code change could be accounting for the speed-up.

kimrutherford:
> The only problem is that the metagenotype names aren't updated in the JSON export

Hi James.

Do you know if the metagenotype display name will be used downstream?  If not then the quickest fix is to remove rangeDisplayName from the JSON output.  All of the metagenotype details are encoded in the genotype and allele details elsewhere in the JSON.  If the JSON file is loaded into a database with a normalised schema perhaps including the metagenotype display name is duplication?

jseager7:
> Do you know if the metagenotype display name will be used downstream?

Unfortunately, I suspect it is being used by one of our collaborators, purely for the sake of convenience. I agree it's redundant, but it could be a real pain to expect them to change their parsing to generate the display names dynamically. They're using a JSON database, so perhaps normalisation is less of an issue.

I'm happy to have a go at making the metagenotype display name appear correctly; I assume that the only code that would need changing is the `json` subroutine in `Serialise.pm`. Otherwise, we could just remove the display name and I can patch it back in afterwards with a script.

kimrutherford:
OK, no problem.  I was half was through implementing the fix when I thought to ask the question.

kimrutherford:
I've made those changes.  Could you give it a test when you get a chance?

jseager7 (approved):

jseager7:
Seems like the metagenotype display names export correctly now. Thanks very much for your help. I'll merge this now.
```json
{
   "rangeDisplayName" : "lasI_test(wild type)[WT level] Pseudomonas aeruginosa (PA14) / wild type Mus musculus (BALB/c)",
   "rangeType" : "Metagenotype",
   "rangeValue" : "11fa7733063e791c-metagenotype-24",
   "relation" : "compared_to_control"
}
```

Issue #2268, opened 2020-04-21
Add admin script option for renaming strains


kimrutherford:
```
canto_admin.pl --rename-strain old-name new-name
```

jseager7:
Do we have a plan for what happens when a strain is renamed to match a strain that already exists in the database? Will the script just fail?

I was wondering – assuming we've got no way to delete strains at the moment – whether it would be better to allow strains to be merged when they're renamed to the same name (in case we discover in future that a strain that we have categorised as a distinct strain turns out to be a synonym of an existing strain). We can afford to wait until that happens, though.

kimrutherford:
Thanks for that.  I'll make sure that the script checks for any existing strain with the same name before renaming.

kimrutherford:
Hi James.

I've added a new option to the `canto_admin.pl` script:

```
  ./script/canto_admin.pl --rename-strain taxonid old_name new_name
```

The script will report an error if there is an existing strains with the new strain name.

Let me know if it works for you.  Could you close the issue if it looks OK?

jseager7:
@kimrutherford Thanks for this. I've tested the script and it works fine for ASCII characters. Unfortunately, some of our strains have UTF-8 characters in them, and entering them from the command line is convoluted. If you're renaming _from_ UTF-8 to ASCII, it might be alright (haven't tested that), but introducing UTF-8 characters in the new strain name doesn't seem to be handled correctly.

With the following contrived example, renaming 'PH-1' to 'PH-ひ', I tried escaping the UTF-8 character in the prompt, which allowed it to load:

```sh
./canto/script/canto_docker /canto/script/canto_admin.pl \ 
--rename-strain 5518 PH-1 PH-"$(printf '\xe3\x81\xb2')"
```

But once it gets into Canto, the encoding gets mangled, displaying as 'PH-ã²':

![image](https://user-images.githubusercontent.com/37659591/82216654-2c99eb80-9911-11ea-9dc2-e037e548ffed.png)

Maybe the Perl script isn't expecting to parse command line arguments as UTF-8?

My terminal locale is definitely UTF-8, so the escape sequence should be correct.

```sh
vagrant@canto-vm:~/canto-docker$ echo $LANG
en_US.UTF-8
```
Even if we fixed the script, I'm not sure if having to interpolate `printf` calls into the command to insert UTF-8 characters is a good idea. Would it be possible to supply the arguments from a file instead in these cases?

kimrutherford:
Hi James.

[...]

PR #2566, opened 2022-04-20
Fix canto_admin.pl not handling Unicode characters in arguments

jseager7:
It looks like commit a8e2c7edee4e77bca8df582f763fc80c64a7dc05 caused a regression in `canto_admin.pl` by removing the `-CA` flags for Perl in the shebang line:
```diff
- #!/usr/bin/perl -CA
+ #!/usr/bin/env perl
```
These flags were being used to allow the script to process Unicode characters in its arguments, which PHI-Canto needed for the `--rename-strain` and `--merge-strains` options (see https://github.com/pombase/canto/issues/2268). I'm assuming that the change in the shebang line is in some way connected to the fix for the performance issues in the transaction, so I've come up with a different fix that uses the `Encode` package to decode the arguments to UTF-8 when they're parsed in the relevant subroutines:
```Perl
  if (defined $merge_strains) {
    my $taxonid = shift @ARGV;
    my $old_name = decode("utf-8", shift @ARGV);
    my $new_name = decode("utf-8", shift @ARGV);
```
I'm presuming this fix assumes that the arguments are being entered as UTF-8, so it might not work on systems where the encoding of the calling program is something else. I found a [more sophisticated fix](https://stackoverflow.com/a/2037520) that uses the `I18N::Langinfo` package to read the locale from the outside data source, but it didn't work for me (it just seemed to corrupt the Unicode characters instead).

@kimrutherford If you can think of any better fix for this, feel free to close this PR.

### DIFF
--- a/script/canto_admin.pl
+++ b/script/canto_admin.pl
@@ -9,6 +9,7 @@ use Try::Tiny;
 
 use File::Basename;
 use Getopt::Long;
+use Encode qw(decode);
 
 BEGIN {
   my $script_name = basename $0;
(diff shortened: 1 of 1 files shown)


kimrutherford:
Sorry about that!  Thanks for the fix.

Issue #2598, opened 2022-07-21
Add curators field to JSON export schema


jseager7:
Issue https://github.com/pombase/canto/issues/2594 added a new `curators` field to the JSON export, so `etc/export.schema.json` will need to be updated to match.

kimrutherford:
Hi James.  Just so you know, you won't see the new `annotation_curators` field in your JSON output because it's only present when the curator names are exported using `--export-curator-names`.

jseager7:
Thanks. I'll update the schema anyway since I use it more for documentation than validation, even for our own Canto instance.

Issue #2594, opened 2022-07-02
Add a "curators" field to JSON output and change session attribution algorithm


kimrutherford:
We should give attribution to all curators who have made an annotation in a session and export the details of all those users into a new "curators" field in the JSON file

From pombase/website#525:

  - ignore who the session who assigned to
  - attribute those people who have made at least one annotation
  - if any of the annotations are by community curators then call it a community session

Needed for APICURON.

jseager7:
This shouldn't cause any problems for our pipeline, since currently we don't include curator names on our website. I know that we _should_ credit community curators, but I'm pretty sure this use of personal data isn't covered by our privacy policy.

If we decide to credit curators, we might need some way to allow them to opt out of being displayed on the website (ideally when they start their curation session in Canto), even though I think it's unlikely they'd want to curate anonymously. Have you considered this at PomBase?

Currently we exclude the `--all` flag when exporting as JSON so we don't accidentally expose any of the curators' personal data. It would be good if the new `curators` property could also be hidden by default.

kimrutherford:
> This shouldn't cause any problems for our pipeline, since currently we don't include curator names on our website.

OK, thanks.

> If we decide to credit curators, we might need some way to allow them to opt out of being displayed on the website (ideally when they start their curation session in Canto), even though I think it's unlikely they'd want to curate anonymously. Have you considered this at PomBase?

We figure that if they're contributing annotation for their own paper, they won't mind being credited with the annotation.  But in any case we only show the curator names on the publication pages that have the author list (like https://www.pombase.org/reference/PMID:34958661) so we're not displaying anything extra.

If a community curator did want to contribute anonymously they could contact our Helpdesk, but as far as I know that's never happened.

We have a privacy policy in case you're curious: https://www.pombase.org/about/privacy-policy

kimrutherford:
I've added a "annotation_curators" field to the JSON file exported from Canto.  It looks like this:

```json
"annotation_curators" : [
   {
      "annotation_count" : 22,
      "community_curator" : true,
      "name" : "Yasushi Hiraoka"
   },
   {
      "annotation_count" : 12,
      "community_curator" : false,
      "name" : "Valerie Wood"
   }
]
```

It's called "annotation_curators" to make it clear that it contains the curators of the annotations rather than a user who the session was assigned to.  Sometimes the assigned curators do no curation.

I've also added a new flag in the JSON file "has_community_curation" which is set to true if any of the annotations were created by a community curator.

[...]

PR #2606, opened 2022-08-15
Update export JSON schema with new curators fields

jseager7:
(Closes https://github.com/pombase/canto/issues/2598)

This PR adds the `annotation_curators` and `has_community_curation` properties to the JSON schema for the Canto JSON export. The properties were originally added to the export by https://github.com/pombase/canto/issues/2594.

- - -

@kimrutherford The new schema could be made a bit stricter by adding `required` properties, but I've erred on the side of caution since I don't know what's guaranteed to be exported.

Can you just confirm the following assumptions? If so, I could add some more constraints.

* If the curation session has no annotations, the `annotation_curators` array is exported as an empty array.
* The `has_community_curation` flag is always exported, and is `false` if the session has no annotations.
* If the `annotation_curators` array is not empty, then all the properties of the objects in the array must always be present (meaning all four of the currently implemented object properties are required).
* If the `annotation_curators` array is not empty, then `annotation_count` must be greater than zero (presumably curators are not included here if they have made no annotations).

Note that all of the above assumes the export is exported with the `--export-curator-names` flag set.

Also, for the purposes of determining who's a community curator: are curators logged in as admins the only users who are _not_ classified as community curators?

### DIFF
--- a/etc/export.schema.json
+++ b/etc/export.schema.json
@@ -492,6 +492,46 @@
                   "description": "Boolean string that tracks whether the curation session is in admin mode. For internal use only.",
                   "$ref": "#/definitions/stringInteger"
                 },
+                "annotation_curators": {
+                  "title": "Annotation curators",
+                  "description": "Tracks statistics about the annotations made by all curators involved in the curation session.",
+                  "type": "array",
+                  "items": {
+                    "title": "Annotation curator details",
+                    "description": "Statistics about the annotations made by a curator involved in the curation session.",
+                    "type": "object",
+                    "properties": {
+                      "annotation_count": {
+                        "title": "Annotation count",
+                        "description": "The number of annotations made by the curator in the curation session.",
+                        "type": "integer",
+                        "minimum": 1
+                      },
+                      "community_curator": {
+                        "title": "Community curator",
+                        "description": "True if the curator who made the annotation was a community curator (that is, not a member of the admin curation team).",
+                        "type": "boolean"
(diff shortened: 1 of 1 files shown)


kimrutherford:
> If the curation session has no annotations, the annotation_curators array is exported as an empty array.

Yep!

> The has_community_curation flag is always exported, and is false if the session has no annotations.

Yep!

> If the annotation_curators array is not empty, then all the properties of the objects in the array must always be present (meaning all four of the currently implemented object properties are required).

Yep!  Although the "orcid" field can be null.

> If the annotation_curators array is not empty, then annotation_count must be greater than zero (presumably curators are not included here if they have made no annotations).

Yep!

If the `--export-curator-names` flag is used, `has_community_curation` is still written and `annotation_curators` will be absent.

> Also, for the purposes of determining who's a community curator: are curators logged in as admins the only users who are not classified as community curators?

The community curators are those whose role is "user" in the "person" table.  If you go to the something like "https://canto.phi-base.org//view/list/person?model=track&order_by=%3Erole" you'll be able to see the admin vs user roles.  Logging in isn't necessary to have annotation counted as admin 
vs user.  The only thing that matters is who the session was assigned to when the annotation was made.

jseager7:
Thanks for the feedback. I'll set the relevant fields to required in the schema.

> The community curators are those whose role is "user" in the "person" table.

Oh yeah, I forgot that the role is set when the user is added to the database.

Issue #2610, opened 2022-08-24
GI overexpression prompts


manulera:
@ValWood @kimrutherford 

The full list of constrains is in this issue: https://github.com/pombase/canto/issues/2593

I think the prompts from Dosage Lethality and Dosage Growth defect would be better if they looked like the prompts in parenthesis here:

![Screenshot 2022-08-24 at 12 16 05](https://user-images.githubusercontent.com/22526102/186405228-eb3bba23-5e67-4962-a94c-10d19edece03.png)

Instead of:
![Screenshot 2022-08-24 at 12 13 25](https://user-images.githubusercontent.com/22526102/186405333-aaaeb360-c3e7-4294-97e6-60e3ac170afa.png)
![Screenshot 2022-08-24 at 12 12 03](https://user-images.githubusercontent.com/22526102/186405338-5fedbb19-3eb2-4591-b516-f7465008aa24.png)

I think they make the directionality more clear. Perhaps the `overexpressed` last word can be removed (to say `inviable in pressence of pkl1[overexpression]` instead of `inviable in pressence of overexpressed pkl1[overexpression]`)

The one from Dosage Rescue is OK, but I would still remind the user that the phenotypes must be growth related, so `lethality or growth defect phenotypes rescued by...`. They may select dosage rescue instead of phenotypic suppression and might be confused not to find the desired phenotypes in there.

![Screenshot 2022-08-24 at 12 12 21](https://user-images.githubusercontent.com/22526102/186405335-863b17e9-64d4-4490-8faf-02dd3a118ccb.png)

ValWood:
The prompt suggestions look good.

kimrutherford:
> I think the prompts from Dosage Lethality and Dosage Growth defect would be better if they looked like the prompts in parenthesis here:

Can we use all of those prompts in the dialog?  It would be simpler if we could.

kimrutherford:
We had a discussion about this a while ago: https://github.com/pombase/canto/issues/2559#issuecomment-1094835283

And a bit here too: https://github.com/pombase/canto/issues/2577

ValWood:
@manulera  can you check that this text works for all prompts. If so it is OK to change.

kimrutherford:
> Hi @kimrutherford. I had a look at canto.yaml, and I made a PR with the prompts as I think they could be 

Do you think this is OK in the test Canto now?

manulera:
[...]

PR #2625, opened 2022-08-30
changes to canto.yaml added

manulera:
See https://github.com/pombase/canto/issues/2610


kimrutherford:
Hi @manulera 

Thanks for the fixes.  I thought it was best to merge the changes and include them in the test Canto straight away.  It's easier to check that it's all OK.

Issue #1543, opened 2018-07-03
Think about improving UniProt XML parsing for gene names


kimrutherford:
On Tuesday 3 July 2018 at 09:57:47, James Seager wrote:

> There's been some discussion in the following issue about the fact
> that (as I understand it) UniProt sometimes returns some long
> systematic identifiers instead of shorter gene names for some genes: 
>
> https://github.com/pombase/canto/issues/1508#issuecomment-401030625
>
> Val said that "the gene name should be retrieved from UniProt and
> displayed in the annotation tables (it will usually be more
> readable). This gene hasn't yet been assigned a real gene name...."
>
> Is this anything to do with Canto's logic (meaning, is Canto
> sometimes using the wrong data for display purposes), or is it a
> UniProt issue?

I'm not sure.  UniProt does some odd things but we might be able to
improve our XML pasrsing.

Here's the XML from that issue:
  https://www.uniprot.org/uniprot/A0A1Q4I1Y1.xml

And here is a better annotated protein:
  https://www.uniprot.org/uniprot/O74638.xml

In O74638 we get the gene name from this bit:

```xml
  <gene>
    <name type="primary">URA7</name>
    ...
```

The code looks for a gene name with type "primary".

If that bit is missing (like in A0A1Q4I1Y1) we fall back to the name field at the top of the entry ("A0A1Q4I1Y1_9MYCO" in this case):

```xml
  <entry dataset="TrEMBL" created="2017-04-12" modified="2017-06-07" version="3">
    <accession>A0A1Q4I1Y1</accession>
    <name>A0A1Q4I1Y1_9MYCO</name>
    ...
```

Using that name might not be best.  As Val points out there's a better name for A0A1Q4I1Y1.  It's in the `<gene>` section:

```xml
  <gene>
    <name type="ORF" evidence="4">BRW65_00080</name>
  </gene>
```

Maybe we should use that?

A complication is that there can be more than one name with type "ORF"
in that section so we'd just have a pick one and hope for the best.

I think this needs discussion.

barnabynorman:
It should also be noted that gene names being returned for the same organism are sometimes different preventing the creation of a multi alleles.

For example on uniprot we see:
Q00909 -> Gibberella zeae (strain PH-1 / ATCC MYA-4620 / FGSC 9075 / NRRL 31084) (Wheat head blight fungus) (Fusarium graminearum)
P0CS93 -> Gibberella zeae (Wheat head blight fungus) (Fusarium graminearum)

[...]

PR #2675, opened 2022-11-29
Also check for "ordered locus" parsing UniProt XML

kimrutherford:
Some entries have an useful "ordered locus" synonym, but no other synonyms or gene names.

Refs pombase/canto#1543

### DIFF
--- a/lib/Canto/UniProt/UniProtUtil.pm
+++ b/lib/Canto/UniProt/UniProtUtil.pm
@@ -102,6 +102,17 @@ sub parse_results
         for my $synonym (@{$entry->{gene}->[0]->{name}}) {
           my $synonym_content = _content($synonym);
 
+          if ($synonym->{type} eq 'ordered locus') {
+            $name = $synonym_content;
+            last;
+          }
+        }
+      }
+
+      if (!defined $name) {
+        for my $synonym (@{$entry->{gene}->[0]->{name}}) {
+          my $synonym_content = _content($synonym);
+
           if ($synonym->{type} eq 'ORF') {
             $name = $synonym_content;
             last;


lib/Canto/UniProt/UniProtUtil.pm
```diff
@@ -98,6 +98,17 @@ sub parse_results
         }
       }
 
+      if (!defined $name) {
+        for my $synonym (@{$entry->{gene}->[0]->{name}}) {
+          my $synonym_content = _content($synonym);
+
+          if ($synonym->{type} eq 'ordered locus') {
```
jseager7, line 105: Could this check be merged with the check for ORF names below? It seems like the other logic and iteration is redundant. Here's what I mean:

```pl
if ($synonym->{type} eq 'ordered locus' ||  $synonym->{type} eq 'ORF') {
```
The short-circuiting should mean that ordered locus names have higher precedence than ORF names. I've tested this on an accession with both name types ([Q0DHH6](https://www.uniprot.org/uniprotkb/Q0DHH6/entry)) and it uses the ordered locus name (although I did have to refresh the gene cache first for some reason). I also tested an accession with only an ORF name ([A0A1Q4I1Y1](https://www.uniprot.org/uniprotkb/A0A1Q4I1Y1/entry)) and that uses the ORF name instead.

jseager7 (approved):
I've tested this change and it works as expected for the accession number [Q9I1W9](https://www.uniprot.org/uniprotkb/Q9I1W9/entry). Thanks. I've just added one suggestion for refactoring.

lib/Canto/UniProt/UniProtUtil.pm
kimrutherford: I wanted to make sure the "ordered locus" names where given priority if there are cases like this:

```xml
    <gene>
      <name type="primary">POL3</name>
      <name type="ORF">D2366</name>
      <name type="ordered locus">YDL102W</name>
    </gene>
```

If we have just one loop that tests for "ordered locus" and "ORF", for this XML the ORF name will be found first and stored in `$name`.  The  "ordered locus" won't be reached.

Issue #2673, opened 2022-11-28
Using Canto with Docker rootless mode


jseager7:
Out of curiosity, I recently tried switching to Docker [rootless mode](https://docs.docker.com/engine/security/rootless/) on my development machine to see if it worked with Canto.

The good news is, Canto seems to work fine in rootless mode. The bad news is, it requires reverting a previous commit to `canto_start_docker` (da1e4e00aeab1e17f6640778c2df2cdc9535f430) that replaced published ports with host networking:

```sh
CANTO_DOCKER_RUN_ARGS="--rm -ti --publish 5000:5000"  # other arguments omitted
```

@kimrutherford Based on the commit message, you changed to host networking to make it "easier to connect with host services". I haven't tested any networking integration outside of connecting to Canto via a web browser, so I don't know if this change is going to break any other workflows. It might not bode well for running Canto with Vagrant (I should test this). Can you think of anything else that might be complicated by using published ports?

By the way, I can't find any obvious way to check whether the host machine is using Docker in rootless mode, so it's probably not straightforward to add a check for rootless mode and adjust the Docker arguments accordingly.

kimrutherford:
I wish I had added a better commit message when I added host networking.  But I'm fairly sure that the "easier to connect with host services" means connecting to a PostgreSQL server.  FlyBase use Canto with Docker and their Canto needs to connect to an instance of the full FlyBase database.  I was easier with host networking.

If you're interested in using rootless mode for PHI-Canto we could add a new option to `canto_start_docker` to disable host networking and to add a `--publish` flag.

jseager7:
Thanks for clarifying. In that case, adding an option sounds like the best choice. The most straightforward would be a `--rootless` argument that adds `--publish 5000:5000` when set, but since published ports aren't only needed for rootless mode, I'd prefer something more generic.

Maybe an argument named `--port` would be the most obvious, or `--publish` if we want to be consistent with Docker (since `--publish` seems to require specifying both a container port and a host port). If the argument is not set, then we default to host networking.

(Ideally, we would default to published ports since `--network=host` is apparently insecure ([see here](https://docs.docker.com/engine/reference/run/#network-host)), but that would cause problems for FlyBase now.)

I'm happy to create a PR for this if either of the options are okay with you.

kimrutherford:
> Maybe an argument named --port would be the most obvious, or --publish if we want to be consistent with Docker

That makes sense.  Either `--port` or `--publish` sounds OK to me.

[...]

PR #2678, opened 2022-12-01
Add --publish option to canto_start_docker

jseager7:
I've added a `--publish` option to the `canto_start_docker` script which passes its value to the `--publish` option of `docker run`. This provides an alternative to `--network=host` for cases where host networking is not supported or publishing a port is preferred (such as using Docker in rootless mode). The option can also be written as a flag (`-p`) to be consistent with Docker's specification.

If the `--publish` option is not set, then `docker run` defaults to `--network=host` to maintain compatibility with existing scripts.

I rewrote the option parsing to use a `case` command to make it easier to handle flags and long options at the same time. Thankfully the Docker implementation of `--publish` doesn't require an equals to separate the option from its argument, so parsing is simpler.

The script doesn't attempt to verify the port argument because Docker already does that.

I also fixed some indentation, reformatted the script's help text, and brought the help text up to date with the latest set of options.

@kimrutherford I've tested this in rootless mode and it seems to work fine, and I've tested using the default Docker daemon (as root) and I'm still able to connect using host networking.

### DIFF
--- a/script/canto_start_docker
+++ b/script/canto_start_docker
@@ -2,12 +2,13 @@
 
 if [ "$1" = "-h" ]
 then
-    echo "usage:"; echo "   $0 [--debug] [--auto]"
+    echo "usage:"; echo "   $0 [--debug] [--no-tty] [--publish port]"
     echo;
     echo "options:"
-    echo "  --debug: turn on the CANTO_DEBUG environment variable"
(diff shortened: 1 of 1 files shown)


jseager7:
Also, this will close #2673.

kimrutherford:
Thanks James.  It all looks good to me.

PR #2687, opened 2023-01-30
Fix typo in FYPO annotation documentation

### LINKED ISSUE #2686: Amino acid ordering in allele descriptions
@ValWood For allele types that involve an amino acid substitution, is the new amino acid supposed to be placed after the old amino acid, or before?

The PomBase help text has the following instructions:

> For combinations of substitutions, insertions, and/or deletions, separate parts of the description with commas. For example, "858-MKGYP,F124D" means five amino acid residues were inserted after position 858, **and Asp was changed to Phe at position 124**.

The text mentions a substitution from Asp (D) to Phe (F), but the allele description (F124D) seems to imply the opposite direction, from Phe (F) to Asp (D). I've found guidance from one other source ([here](https://varnomen.hgvs.org/recommendations/protein/variant/substitution/)) that recommends placing the old amino acid before the new amino acid:

> Format: “prefix”“amino_acid”“position”“new_amino_acid”, e.g. p.(Arg54Ser)

Is the example in the documentation a mistake, or does PomBase use different conventions?

jseager7:
(Fixes #2686)

The order of amino acids in the example "858-MKGYP,F124D" didn't match the description in the text "Asp was changed to Phe". The example is correct, so I've updated the text to match.

### DIFF
--- a/root/docs/fypo_annotation.mhtml
+++ b/root/docs/fypo_annotation.mhtml
@@ -99,7 +99,7 @@ box will pop up where you can add allele details:
 	For combinations of substitutions, insertions, and/or
 	deletions, separate parts of the description with commas. For
 	example, "858-MKGYP,F124D" means five amino acid residues were
-	inserted after position 858, and Asp was changed to Phe at
+	inserted after position 858, and Phe was changed to Asp at
 	position 124.
       </li>
       <li>


jseager7:
I'm leaving this as a PR instead of committing directly in case Canto's version number needs to be bumped after this change.

kimrutherford:
Thanks James.  I'll merge this and bump the version number.

Issue #2691, opened 2023-02-01
Change example prompts for insertions


manulera:
cc: @ValWood 

Hi @kimrutherford. We discussed with Val that from now on, when annotating aminoacid or nucleotide insertions in alleles, the residue should be indicated. E.g. A123-PLL (insert aminoacids PLL after alanine 123). Until now we were recommending to just indicate the index (123-PLL), but I think going forward is better to go this way to prevent errors from going unnoticed.

We would need to change the prompts in canto (placehold field). I will make a pull request about this.

manulera:
The diff of the pull request

```diff
--- a/canto.yaml
+++ b/canto.yaml
@@ -1455,7 +1455,7 @@ allele_type_list:
     allele_name_required: false
     allow_expression_change: true
     expression_required: true
-    placeholder: e.g. 858-MKGYP
+    placeholder: e.g. A858-MKGYP
   - name: "amino acid substitution(s)"
     export_type: amino_acid_mutation
     show_description: true
@@ -1471,7 +1471,7 @@ allele_type_list:
     allele_name_required: false
     allow_expression_change: true
     expression_required: true
-    placeholder: e.g. 858-MKGYP,F124D
+    placeholder: e.g. A858-MKGYP,F124D
   - name: amino acid insertion and deletion
     export_type: amino_acid_insertion_and_deletion
     show_description: true
@@ -1479,7 +1479,7 @@ allele_type_list:
     allele_name_required: false
     allow_expression_change: true
     expression_required: true
-    placeholder: e.g. 858-MKGYP,100-200
+    placeholder: e.g. A858-MKGYP,100-200
   - name: partial deletion and amino acid change
     export_type: amino_acid_deletion_and_mutation
     show_description: true
@@ -1503,7 +1503,7 @@ allele_type_list:
     allele_name_required: false
     allow_expression_change: true
     expression_required: true
-    placeholder: e.g. 132-AATTT
+    placeholder: e.g. C132-AATTT
   - name: "nucleotide substitution(s)"
     export_type: nucleotide_mutation
     show_description: true
```

PR #2693, opened 2023-02-02
Update amino acid insertion notation in phenotype documentation

jseager7:
(Follow up to #2691)

I noticed that the recommended notation for amino acid insertions was changed in `canto.yaml` (see PR #2692), but the corresponding guidance in `fypo_annotation.mhtml` hadn't been updated to match. This PR fixes that.


kimrutherford:
Thanks James.

Issue #2699, opened 2023-02-28
no options to curate GI


ValWood:
I could not add a GI for this double mutant, there were no menu options available:

<img width="1179" alt="Screenshot 2023-02-28 at 15 05 11" src="https://user-images.githubusercontent.com/7359272/221897998-7333439d-7f4b-4eae-ac7c-8db3fcce5c6a.png">

ValWood:
curs/88f3e63970618c74

ValWood:
expected 
Phenotypic Enhancement

kimrutherford:
It's doing what the configuration says:
  - it's not a population phenotype so the type must be one of:
     - Phenotypic Enhancement
     - Phenotypic Suppression
  - and it's not a double deletion and neither allele is overexpressed so it also must be one of:
    - Synthetic Growth Defect
    - Synthetic Lethality
    - Synthetic Rescue

and the intersection of those lists is empty.

ValWood:
OK, I'll look at this today. I might be making the annotation incorretly.

manulera:
Hi @kimrutherford @ValWood 

I think this interaction should be allowed from what we discussed:

https://www.pombase.org/documentation/genetic-and-physical-interactions

Phenotypic enhancement / suppression should be allowed for any type of alleles (no restriction on allele types). `not_population_evidence_codes` should be the only restriction.

ValWood:
Agreed, sorry for the slowness, working backwards!

kimrutherford:
I think I've fixed this.  Could you check that it's OK?

ValWood:
Yes it works.

2 thoughts:
1. If we select a phenotype that is enhanced or repressed, it seems to be possible to edit the rescued phenotype with no consequences 
2. For phneotypic enhancement it might be a bit strange sometimes to select a "enhanced phenotype"  and/or a direction.
For example if 2 'normal' exhibit a defect when combined. We may need to relax the need to select a direction and a phenotype.

manulera:
> For example if 2 'normal' exhibit a defect when combined. We may need to relax the need to select a direction and a phenotype.

We asked this to BioGrid and they said this in their email:

> Your example of a synthetic phenotype (non-growth or viability) that does not exist in the single mutants is a good example. When we came across examples like this in the literature we decided to use "phenotypic enhancement" as a best fit option. Although it's not ideal as it implies the presence of the phenotype in one of the starting single mutants.

Full response here: https://github.com/pombase/curation/issues/3295

[...]

PR #2707, opened 2023-03-14
Issue 2699

manulera:
New GI proposed in #2699 "Synthetic Phenotype" added to canto.yaml


kimrutherford:
Thanks!  That's great.

Issue #2732, opened 2023-05-30
Bulk annotation formats link is not configurable


jseager7:
The 'bulk annotation formats' link on the gene entry page (root/curs/gene_upload.mhtml) always links to the [page on PomBase](https://www.pombase.org/submit-data/bulk-annotation) regardless of which database Canto is configured to use.

I don't think PHI-base has any use for this link at the moment since we don't (to my knowledge) accept any bulk annotation formats.

@kimrutherford Probably the simplest fix would be to disable this link for any database except PomBase, by checking whether `$config->{database_name}` equals `PomBase`, as it does in your [canto_deploy.yaml](https://github.com/pombase/pombase-config/blob/c73fdd92e30bed0d1940d1fbdbbf3aa2accf10d8/canto/pombase/production_canto_deploy.yaml).

The general fix would be to add a new configuration property to set the URL to a page that describes bulk annotation formats for the database of the Canto instance. If this property does not exist (or is empty), the text for the link is hidden.

Here's the relevant markup:

https://github.com/pombase/canto/blob/2f8fe11c217b52a69251cb589abdf798dab3767b/root/curs/gene_upload.mhtml#L21-L23

PR #2734, opened 2023-06-07
Show PomBase-only link for bulk annotation format docs only for PomBase

kimrutherford:
Refs pombase/canto#2732


kimrutherford:
Hi @jseager7

If you have a chance, could you try this patch to fix #2732?

Thanks.

jseager7 (approved):
@kimrutherford Thanks for making this change. I pushed one more change to hide the following line, since it also relates to bulk annotation and doesn't make sense to still be shown when the first line is hidden:

> Note: Only supply high confidence interactions for large datasets.

kimrutherford (commented):
Thanks for the fix.

PR #2745, opened 2023-07-12
Add a privacy policy notice to the Curator details page

jseager7:
I've added a `privacy_policy_url` configuration option that stores the URL to the privacy policy for the Canto instance. If configured, a message like the following will be shown on the Curator details page before the curators enters their details:

> Your name, email and ORCID ID will be used under the terms of the \[database_name\] privacy policy. By continuing past this page, you agree to the terms of this policy.

Here's how it looks on the webpage:

image

@kimrutherford If you want to apply this to PomBase, here's the configuration:

```yaml
privacy_policy_url: "https://pombase.org/about/privacy-policy"
```

I've also documented the configuration property in `configuration_file.md`.

### DIFF
--- a/root/curs/assign_session.mhtml
+++ b/root/curs/assign_session.mhtml
@@ -18,5 +18,8 @@ $reassign
 %   }
 % }
     <% $form |n %>
+%   if ($c->config()->{privacy_policy_url}) {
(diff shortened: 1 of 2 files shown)


kimrutherford:
> @kimrutherford If you want to apply this to PomBase, here's the configuration:

Thanks, that's a good idea.

And thanks for the PR.

PR #2746, opened 2023-07-12
Autocomplete allele name for transformant allele type

jseager7:
Requested in https://github.com/PHI-base/curation/issues/157#issuecomment-1632586675

I've added configuration to allow Canto to automatically complete allele names for the transformant allele type. The format is the allele name, followed by a space, followed by the string 'transformant'.

@kimrutherford Some points to consider:

* The transformant allele type isn't defined in `canto.yaml`, but I don't think this is a problem since the conditional that applies the autocomplete template will never be true unless the transformant allele type is configured somewhere (and even if the condition somehow becomes true, it shouldn't cause any errors).

* I changed a consecutive `if` statement in `Config.pm` to `elsif` since it looked like the cases were mutually exclusive. You might want to double-check this.

### DIFF
--- a/canto.yaml
+++ b/canto.yaml
@@ -1451,6 +1451,7 @@ implementation_classes:
 
 wildtype_name_template: '@@gene_display_name@@+'
 deletion_name_template: '@@gene_display_name@@delta'
+transformant_name_template: '@@gene_display_name@@ transformant'
 
 allele_type_list:
   - name: wild type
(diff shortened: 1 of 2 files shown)


kimrutherford:
> The transformant allele type isn't defined in canto.yaml, but I don't think this is a problem since the conditional that applies the autocomplete template will never be true unless the transformant allele type is configured somewhere

I agree.  It should be fine.

> I changed a consecutive if statement in Config.pm to elsif since it looked like the cases were mutually exclusive. You might want to double-check this.

Thanks.  It looks good to me and is tidier.  I have a bit of a mental block about `elsif`.  I never think to use it.

Issue #2677, opened 2022-11-30
Allow Canto to replace a UniProtKB accession number


jseager7:
(Requested by @CuzickA, possibly related to #2669)

There's been a few curation sessions now where we've needed to change one UniProtKB accession number to another. Usually this is because the accession number is from the wrong species or because the accession is not from the reference proteome / strain.

Unfortunately, most of these changes are needed _after_ alleles, genotypes and metagenotypes have been added to the session, so it's very tedious to make the change manually.

This could be fixed by something as a simple as a `canto_curs_map` script to update the accession number in the database, but there's the added complication that we have to prevent invalid or obsolete UniProtKB accession numbers from being used, and we have to refresh the cached gene information so that the correct gene name and product is displayed.

If tools for performing bulk renames are under consideration, then it would be really useful to have something that could handle UniProtKB accession numbers. I'm hoping most of the code needed for this is already in `UniProtUtil`. I'd be happy to help with the implementation and testing.

kimrutherford:
> This could be fixed by something as a simple as a canto_curs_map script to update the accession number in the database

Yep, I think it should be straight forward if you know the old and new accessions. 

> there's the added complication that we have to prevent invalid or obsolete UniProtKB accession numbers from being used

Could you explain this bit some more?

jseager7:
> Could you explain this bit some more?

My point is that when adding UniProtKB accession numbers using Canto's web interface, it prevents adding accession numbers that are obsolete (i.e. moved to UniParc) or invalid (don't map to any existing accession).

For example, the following obsolete accession number shows a warning message in the web interface: 

No genes found for these identifiers: I1RA15

But if the script doesn't query the new accessions against UniProtKB before doing the replacement, then it will be possible to replace a valid accession with an obsolete or non-existent accession, and I don't know whether that will cause problems for the code.

Maybe `UniProtUtil` can already handle this case by showing only the accession number and leaving the other fields blank. We'd need this case handled anyway, since even UniProtKB accessions that we've already may be at risk of going obsolete in the future.

An initial version of the script could just replace the accession number without any checks, and we might not need the checks at all if Canto doesn't break when an invalid accession number is entered, but it would be nice to have since curators entering the 'wrong' accession number could be an unfortunately common occurrence.

kimrutherford:
Thanks James.  It should be no problem to call the function in UniProtUtil that gets entry details for the new accession before making changes.

Let's have a Zoom chat about this sometime.  Maybe at the start of January?

[...]

PR #2822, opened 2024-04-05
Add start of --change-gene-id admin action

kimrutherford:
Refs pombase/canto#2677

### DIFF
--- a/lib/Canto/Track/TrackUtil.pm
+++ b/lib/Canto/Track/TrackUtil.pm
@@ -498,6 +498,36 @@ sub change_gene_id
         $allele->update();
       }
     }
+    
+    my $annotation_rs = $cursdb->resultset('Annotation');
+
+    while (defined (my $annotation = $annotation_rs->next())) {
+      my $data = $annotation->data();
+      my $changed = 0;
+      
+      my $extension = $data->{extension};
+      
+      if (defined $extension) {
+        map {
+          my $orPart = $_;
+          map {
+            my $andPart = $_;
+            if ($andPart->{rangeType} && $andPart->{rangeType} eq 'Gene') {
+              if ($andPart->{rangeValue} eq $from_id) {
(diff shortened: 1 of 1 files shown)


jseager7:
@kimrutherford I've added code to update the gene name in the allele name, and to update the gene name (`rangeDisplayName`) and primary identifier (`rangeValue`) in annotation extensions.

I've tested this with UniProtKB accession numbers in PHI-Canto and it seems to work fine, but you'll have to test it with PomBase / Chado as the gene source because I don't know whether my changes are safe to use with those.

Specifically, I'm assuming there will only ever be one result returned from UniProtKB in the `$from_id_lookup_result` and `$to_id_lookup_result`, which allows me to simplify the code for getting the old and new gene names to this:
```perl
  my $old_name = $from_id_lookup_result->{found}->[0]->{primary_name};
  my $new_name = $to_id_lookup_result->{found}->[0]->{primary_name};
```
But I don't know whether that assumption holds for PomBase, or if it even holds for UniProtKB. I guess a safer solution would be to iterate through the results and find the first result where the primary identifier matches the `$from_id` (then do the same for `$to_id`), then get the new gene name from that result. I couldn't figure out how to do this at the time though.

kimrutherford:
Hi James.

Thanks very much for those changes.  It all looks good to me.  I'm going to merge the PR.  We can added any fixes to the main branch.

> But I don't know whether that assumption holds for PomBase, or if it even holds for UniProtKB. 

It's not going to be a problem for PomBase because our lookup code will only return a single gene for a systematic ID.

I think it's OK for UniProt too since we're looking up accessions.

The web service used for the UniProt lookup is configured in `canto.yaml`:

```yaml
webservices:
  uniprot_batch_lookup_url: 'https://rest.uniprot.org/uniprotkb/search?format=xml&query='
```

jseager7:
> It all looks good to me. I'm going to merge the PR.

Great, thanks again for your help.

PR #2830, opened 2024-05-02
Toggle renaming of gene names in alleles in --change-gene-id

jseager7:
I've added a flag `--no-rename` to `canto_admin.pl` that, when set with `--change-gene-id`, skips the renaming of gene names in allele names. This was needed because PHI-base wanted to keep using the gene name that was specified in the publication for our annotations.

The `--no-rename` flag defaults to `0` instead of `undef`. If the `--no-rename` flag is set with any other flag, the script warns and exits:
```
Error: --no-rename is only allowed with --change-gene-id
```
This required adding an argument `$no_rename` to `change_gene_id` in TrackUtil.pm that skips the allele renaming when set. I didn't bother to set a default value for this argument because `change_gene_id` is currently only used in one place.

I've tested this with PHI-Canto where it works fine both with and without the new flag. I haven't tested  any other modes.

### DIFF
--- a/lib/Canto/Track/TrackUtil.pm
+++ b/lib/Canto/Track/TrackUtil.pm
@@ -432,7 +432,8 @@ sub update_annotation_curators
  Function: Change a gene ID (primary_identifier) in every session.
            Also change allele IDs containing the $from_id.
  Args    : $from_id - an existing primary_identifier
-           $to_id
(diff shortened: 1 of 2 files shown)


kimrutherford:
Thanks James.  Both of those patches look great.

